PR #2, opened 2020-07-13
Implement Simulator with approach1

ckyiricky:
Implement simulator structure as shown before.
Implement approach 1 (shared global endpointslices algorithm)

1.Add comments to the code
2.fixed two bugs:
  a. calculation of number of endpointslices 
  b. approach 1 now returns error when input zones have 0 nodes/pods

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import "github.com/googleinterns/k8s-topology-simulator/modeling"
+
+func main() {
+	zoneA := modeling.Zone{Nodes: 30, Endpoints: 60, Name: "a"}
+	zoneB := modeling.Zone{Nodes: 35, Endpoints: 70, Name: "b"}
+	zoneC := modeling.Zone{Nodes: 50, Endpoints: 80, Name: "c"}
+	zones := []modeling.Zone{zoneA, zoneB, zoneC}
+
+	alg, err := modeling.CreateAlg(0.4, 100)
+	if err != nil {
+		panic(err)
+	}
+
+	var sim modeling.TheoreticalSimulator
+
+	// TODO: operations on metrics rather than printing only
+	model, err := modeling.NewModel(zones, alg, sim)
+	if err != nil {
+		panic(err)
+	}
+
+	err = model.StartSimulation()
+	if err != nil {
+		panic(err)
+	}
+	err = model.PrintLastResult()
+	if err != nil {
+		panic(err)
+	}
+}
--- a/modeling/data.go
+++ b/modeling/data.go
@@ -1,0 +1,112 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package modeling
+
+import "errors"
+
+// Zone abstracts the conception of 'zone' in clouds
+type Zone struct {
+	// Nodes is the numer of nodes of this zone
+	Nodes int
+	// Endpoints is the Number of endpoints in this zone
+	Endpoints int
+	// Name of this zone
+	Name string
+	// Number of pods / total number of pods of all zones
+	endpointsRatio float64
+	// Number of nodes / total number of nodes of all zones
+	nodesRatio float64
+}
+
+type regionInfo struct {
+	// Total number of nodes of all zones
+	totalNodes int
+	// Total number of endpoints of all zones
+	totalEndpoints int
+	// Detailed information of each zone, zone name - exact zone
+	zoneDetails map[string]Zone
+}
+
+// EndpointSliceGroup represents all the EndpointSlices under a same label, one
+// group may be made up by many EndpointSlices (when the number of endpoints
+// excceeds the capacity of one EndpointSlice). Since for now there is no need
+// to know how the group is composed, we keep them as a whole
+type EndpointSliceGroup struct {
+	// Label is a unique identifier for an EndpointSliceGroup. This often
+	// represents a topology label that the group will be consumed from.
+	Label string
+	// Composition stores contribution of endpoints in this group from different
+	// zones
+	Composition map[string]int
+	// Weights of endpoints in this group for different zones while routing
+	Weights map[string]float64
+}
+
+// Traffic represents the detailed traffic infomation of a zone
+type Traffic struct {
+	// ZoneName of a specific zone
+	ZoneName string
+	// Incoming traffic that the same zone received
+	Incoming float64
+	// Outgoing traffic distribution of the same zone
+	Outgoing map[string]float64
+}
+
+// SimulationResult is to collect metrics of a simulation result
+type SimulationResult struct {
+	// InZoneTraffic is the total ratio of traffic that stays in the same zone
+	InZoneTraffic float64
+	// TrafficDetail groups Traffic by zone name
+	TrafficDetail map[string]Traffic
+	// Workload balance for different zones -- ratio of incoming traffic / ratio
+	// of capacity
+	Workload map[string]float64
+}
+
+// Helper function to calculate number of endpoints of a specific
+// EndpointSliceGroup
+func (e EndpointSliceGroup) numberOfPods() int {
+	total := 0
+	for _, pods := range e.Composition {
+		total += pods
+	}
+	return total
+}
+
+func createRegionInfo(zones []Zone) (regionInfo, error) {
+	if len(zones) == 0 {
+		return regionInfo{}, errors.New("creating zoneinfos with zero length []Zone")
+	}
+	var totalPods, totalNodes int
+
+	region := regionInfo{zoneDetails: make(map[string]Zone)}
+	for _, zone := range zones {
+		if zone.Endpoints <= 0 || zone.Nodes <= 0 {
+			return regionInfo{}, errors.New("invalid zones with number of nodes or endpoints <= 0")
+		}
+		totalPods += zone.Endpoints
+		totalNodes += zone.Nodes
+	}
+	region.totalEndpoints = totalPods
+	region.totalNodes = totalNodes
+	for _, zone := range zones {
+		zone.endpointsRatio = float64(zone.Endpoints) / float64(totalPods)
+		zone.nodesRatio = float64(zone.Nodes) / float64(totalNodes)
+		region.zoneDetails[zone.Name] = zone
+	}
+	return region, nil
+}
--- a/modeling/interface.go
+++ b/modeling/interface.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package modeling
+
+// RoutingAlgorithm interface for different routing algorithms
+type RoutingAlgorithm interface {
+	// CreateSliceGroups translates regionInfo into EndpointSliceGroups
+	CreateSliceGroups(regionInfo) (map[string]EndpointSliceGroup, error)
+}
+
+// TrafficSimulator interface for different simulators
+type TrafficSimulator interface {
+	// Simulate simulates traffic with the provided regionInfo and
+	// EndpointSliceGroups and returns a SimulationResult.
+	Simulate(regionInfo, map[string]EndpointSliceGroup) (SimulationResult, error)
+}
--- a/modeling/model.go
+++ b/modeling/model.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package modeling
+
+import (
+	"errors"
+	"fmt"
+)
+
+// Model wrapper class for the simulation components
+type Model struct {
+	slices    map[string]EndpointSliceGroup
+	alg       RoutingAlgorithm
+	simulator TrafficSimulator
+
+	Region  regionInfo
+	Results []SimulationResult
+	// SliceCapacity is the number of max endpoints per slice
+	SliceCapacity uint
+}
+
+// NewModel creates a model with zones, routing algorithm and traffic simulator
+// and uses the algorithm to create the EndpointSliceGroups
+func NewModel(zones []Zone, alg RoutingAlgorithm, sim TrafficSimulator) (*Model, error) {
+	region, err := createRegionInfo(zones)
+	if err != nil {
+		return nil, err
+	}
+	if alg == nil || sim == nil {
+		return nil, errors.New("can't create model with nil algorithm or simulator")
+	}
+	model := &Model{
+		Region:        region,
+		SliceCapacity: 100,
+		alg:           alg,
+		simulator:     sim,
+	}
+	slices, err := model.alg.CreateSliceGroups(model.Region)
+	if err != nil {
+		return nil, err
+	}
+	model.slices = slices
+	return model, nil
+}
+
+// StartSimulation based on the zones(Region) and EndpointSliceGroups
+func (m *Model) StartSimulation() error {
+	simResult, err := m.simulator.Simulate(m.Region, m.slices)
+	if err != nil {
+		return err
+	}
+	m.Results = append(m.Results, simResult)
+	return nil
+}
+
+// PrintLastResult prints the summary of the last simulation result
+func (m *Model) PrintLastResult() error {
+	numberOfSlices := m.GetNumberOfEndpointSlices()
+	lastResult := m.Results[len(m.Results)-1]
+	fmt.Printf("%% in-zone traffic \t %.2f%%\n", lastResult.InZoneTraffic*100)
+	fmt.Printf("# of endpoint slices\t %v\n", numberOfSlices)
+	fmt.Printf("# of endpoints\t %d\n", m.Region.totalEndpoints)
+	fmt.Println("----------------------------------------------")
+
+	for zone, traffic := range lastResult.TrafficDetail {
+		fmt.Printf("Total to %s \t %.f%% \n", zone, traffic.Incoming*100)
+	}
+	fmt.Println("----------------------------------------------")
+
+	for zone, traffic := range lastResult.TrafficDetail {
+		fmt.Printf("From %s : \n", zone)
+		for z, t := range traffic.Outgoing {
+			fmt.Printf("\t to %s : %.2f%% \n", z, t*100)
+		}
+	}
+	fmt.Println("----------------------------------------------")
+
+	for zone, workload := range lastResult.Workload {
+		fmt.Printf("Workload for %s : \t %.2f%% \n", zone, workload*100)
+	}
+	return nil
+}
+
+// GetEndpointSliceGroups returns the EndpointSliceGroups, don't allow users to
+// manually change this field -- use this getter to get the value
+func (m *Model) GetEndpointSliceGroups() map[string]EndpointSliceGroup {
+	return m.slices
+}
+
+// GetNumberOfEndpointSlices returns the number of EndpointSlices
+func (m *Model) GetNumberOfEndpointSlices() uint {
+	totalSlices := uint(0)
+	for _, slice := range m.slices {
+		pods := slice.numberOfPods()
+		totalSlices += uint(pods) / m.SliceCapacity
+		if uint(pods)%m.SliceCapacity != 0 {
+			totalSlices++
+		}
+	}
+	return totalSlices
+}
--- a/modeling/shared-global-algorithm.go
+++ b/modeling/shared-global-algorithm.go
@@ -1,0 +1,101 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package modeling
+
+import (
+	"errors"
+	"math"
+)
+
+// SharedGlobalAlgorithm takes multiple zones as input and output
+// EntpointSliceGroups composition based on their nodes and endpoints:
+// 1. EndpointSlices will be considered global by default
+// 2. Once the number of endpoints in a zone reaches a specified threshold,
+//	  zone-specific EndpointSlices will begin to be produced.
+// 3. Kube Proxy will consume both global and zone specific EndpointSlices.
+//	  Endpoints in the same zone will be given a higher weight.
+type SharedGlobalAlgorithm struct {
+	// Weight of global EndpointSliceGroup
+	globalWeight float64
+	// Threshold of global EndpointSliceGroup that if the total number of endpoints
+	// <= threshold, all endpoints go to global EndpointSliceGroup
+	globalThreshold int
+}
+
+// CreateSliceGroups takes a region of zones as input and output
+// EndpointSliceGroups
+func (alg SharedGlobalAlgorithm) CreateSliceGroups(region regionInfo) (map[string]EndpointSliceGroup, error) {
+	if region.zoneDetails == nil {
+		return nil, errors.New("can't create EndpointSlices with 0 number of zone")
+	}
+	// The deviation for the traffic and capacity above
+	deviation := make(map[string]float64)
+
+	for _, zone := range region.zoneDetails {
+		// Calculate the deviation based on the capacity(endpoints) and
+		// traffic(nodes) ratio
+		deviation[zone.Name] = float64(zone.Endpoints) - float64(region.totalEndpoints)*zone.nodesRatio
+	}
+
+	// Output EndpointSlices
+	sliceGroups := make(map[string]EndpointSliceGroup)
+	// The global sliceGroup -- might be split into many small global slices
+	// when the number of endpoints > required number of endpoints per
+	// EndpointSlice, i.e. 100 for default. Not be able to divide the big one
+	// into smaller ones that the contributions may vary and there is no need to
+	// do so either.
+	var globalSliceGroup EndpointSliceGroup
+	globalSliceGroup.Label = "global"
+	globalSliceGroup.Composition = make(map[string]int)
+	globalSliceGroup.Weights = make(map[string]float64)
+	for name, zone := range region.zoneDetails {
+		var globalEndpoints int
+		// If total pods <= threshold, all pods go to global slice
+		if region.totalEndpoints <= alg.globalThreshold {
+			globalEndpoints = zone.Endpoints
+		} else {
+			// Otherwise calculate the global contribution of current zone based
+			// on the global weight and the deviation of this zone
+			globalEndpoints = int(math.Min(math.Max(0.0, deviation[name])/alg.globalWeight, float64(zone.Endpoints)))
+		}
+
+		globalSliceGroup.Composition[name] = globalEndpoints
+		globalSliceGroup.Weights[name] = alg.globalWeight
+
+		// Calculate how many endpoints remain in the local zone
+		var localGroup EndpointSliceGroup
+		localGroup.Label = name
+		localGroup.Composition = make(map[string]int)
+		localGroup.Weights = make(map[string]float64)
+		localGroup.Composition[name] = zone.Endpoints - globalEndpoints
+		localGroup.Weights[name] = 1.0
+
+		sliceGroups[name] = localGroup
+	}
+	sliceGroups[globalSliceGroup.Label] = globalSliceGroup
+	return sliceGroups, nil
+}
+
+// CreateAlg -- the constructor of the algorithm , set weight and threshold of
+// the global EndpointSliceGroup
+func CreateAlg(weight float64, threshold int) (SharedGlobalAlgorithm, error) {
+	if weight < 0 || threshold < 0 {
+		return SharedGlobalAlgorithm{}, errors.New("invalid weight/threshold values to init algorithm")
+	}
+	alg := SharedGlobalAlgorithm{globalWeight: weight, globalThreshold: threshold}
+	return alg, nil
+}
--- a/modeling/theoretical-simulator.go
+++ b/modeling/theoretical-simulator.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package modeling
+
+import "errors"
+
+// TheoreticalSimulator calculates the theoretical probability of the traffic
+// distribution
+type TheoreticalSimulator struct{}
+
+// Simulate calculates the theoretical distribution of the traffic
+func (sim TheoreticalSimulator) Simulate(region regionInfo, endpointSlices map[string]EndpointSliceGroup) (SimulationResult, error) {
+	if len(region.zoneDetails) == 0 || len(endpointSlices) == 0 {
+		return SimulationResult{}, errors.New("can't evaluate probability based on empty zones or endpointslices")
+	}
+	// kube-proxy calculation, zones name - endpoints in potential destination sclies
+	kubeProxy := make(map[string]map[string]float64)
+	for name := range region.zoneDetails {
+		kubeProxy[name] = map[string]float64{}
+		for _, slice := range endpointSlices {
+			for compZone, pods := range slice.Composition {
+				kubeProxy[name][compZone] += float64(pods) * slice.Weights[name]
+			}
+		}
+	}
+	// probability of traffic from a zone to other zones, outgoing zone --
+	// probability of going to other zones
+	zoneEndpoints := map[string]map[string]float64{}
+	for zoneName, endpointsDistribution := range kubeProxy {
+		zoneTotal := 0.0
+		for _, endpoints := range endpointsDistribution {
+			zoneTotal += endpoints
+		}
+		zoneEndpoints[zoneName] = map[string]float64{}
+		for destZoneName, endpoints := range endpointsDistribution {
+			zoneEndpoints[zoneName][destZoneName] = endpoints / zoneTotal
+		}
+	}
+
+	// calculate result of one simulation
+	var simResult SimulationResult
+	simResult.TrafficDetail = map[string]Traffic{}
+	simResult.Workload = map[string]float64{}
+	for zoneName, zone := range region.zoneDetails {
+		// zoneX -> zoneX forms inzone traffic
+		simResult.InZoneTraffic += zone.nodesRatio * zoneEndpoints[zoneName][zoneName]
+
+		var traffic Traffic
+		traffic.ZoneName = zoneName
+		// Outgoing traffic distribution
+		traffic.Outgoing = zoneEndpoints[zoneName]
+		for originZoneName, originZone := range region.zoneDetails {
+			// Accumulate total incoming traffic to zoneName
+			traffic.Incoming += originZone.nodesRatio * zoneEndpoints[originZoneName][zoneName]
+		}
+		simResult.Workload[zoneName] = traffic.Incoming / zone.endpointsRatio
+		simResult.TrafficDetail[zoneName] = traffic
+	}
+	return simResult, nil
+}
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/googleinterns/k8s-topology-simulator
+
+go 1.14


main.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+import "modeling/modeling"
+
+func main() {
+	zoneA := modeling.Zone{Nodes: 30, Endpoints: 60, Name: "a"}
+	zoneB := modeling.Zone{Nodes: 35, Endpoints: 70, Name: "b"}
+	zoneC := modeling.Zone{Nodes: 50, Endpoints: 80, Name: "c"}
+	zones := []modeling.Zone{zoneA, zoneB, zoneC}
```
robscott, line 25: I think it might be good to find a way to automate random inputs here in the future.

main.go
```diff
@@ -0,0 +1,17 @@
...
+
+	alg, _ := modeling.CreateAlg(0.4, 100)
```
robscott: Probably want to handle this potential error.

main.go
```diff
@@ -0,0 +1,17 @@
...
+	sim := modeling.CreateDefaultSim(100000)
+
+	mo, _ := modeling.NewModel(zones, alg, sim)
+	mo.StartSimulation()
+	mo.PrintLastResult()
```
robscott: Could we do some kind of averaging/totals/calculations here instead of just printing the last result?

main.go
```diff
@@ -0,0 +1,17 @@
...
+func main() {
+	zoneA := modeling.Zone{Nodes: 30, Endpoints: 60, Name: "a"}
+	zoneB := modeling.Zone{Nodes: 35, Endpoints: 70, Name: "b"}
+	zoneC := modeling.Zone{Nodes: 50, Endpoints: 80, Name: "c"}
+	zones := []modeling.Zone{zoneA, zoneB, zoneC}
+
+	alg, _ := modeling.CreateAlg(0.4, 100)
+	sim := modeling.CreateDefaultSim(100000)
+
+	mo, _ := modeling.NewModel(zones, alg, sim)
```
robscott: Should probably handle error here as well.

modeling/data.go
```diff
@@ -0,0 +1,33 @@
+package modeling
+
+type Zone struct {
+	Nodes     int
+	Endpoints int
+	Name      string
+}
+
+type Endpointslice struct {
```
robscott: nit: EndpointSlice

modeling/data.go
```diff
@@ -0,0 +1,33 @@
+package modeling
+
+type Zone struct {
```
robscott, line 22: I think it would be good to have comments above each struct with a bit of an explanation of the data structure.

modeling/data.go
```diff
@@ -0,0 +1,46 @@
...
+	// Number of nodes of this zone
+	Nodes int
+	// Number of endpoitns of this zone
+	Endpoints int
+	// Name of this zone
+	Name string
+}
+
+type Endpointslice struct {
+	// Lable of this endpointslice
```
robscott: Nit: Label

modeling/data.go
```diff
@@ -0,0 +1,46 @@
...
+	Label string
+	// Contribution of endpoints from different zones
+	Composition map[string]int
+	// Weights that this endpointslice for different zones while routing
```
robscott: This comment is confusing to me.

modeling/data.go
```diff
@@ -0,0 +1,46 @@
...
+	Weights map[string]float64
+}
+
+type Traffic struct {
+	// Name of a specific zone
```
robscott: Name of a specific zone that traffic is going to or coming from? The other fields here seem to suggest that this is both. Should this just be `ZoneName` or something?

modeling/data.go
```diff
@@ -0,0 +1,46 @@
...
+	// Traffic that the same zone received
+	IncomingTraffic float64
+	// Outgoing traffic distribution of the same zone
+	OutgoingTraffic map[string]float64
+}
+
+type Stat struct {
+	// Total ratio of traffic that stays in the same zone
+	InZoneTraffic float64
+	// Traffic details for different zones
```
robscott: Make it more clear what the string represents here. Maybe `Traffic details by zone name`?

modeling/data.go
```diff
@@ -0,0 +1,46 @@
...
+	Traffic map[string]*Traffic
```
robscott: Unless there are significant performance benefits, I think I'd prefer `map[string]Traffic`

modeling/data.go
```diff
@@ -0,0 +1,46 @@
...
+	// Workload balance for different zones -- ratio of incoming traffic / ratio of capacity
+	Workload map[string]float64
+}
+
+// Helper function to calculte number of endpoints of a specific endpointslice
```
robscott: Nit: calculate

main.go
```diff
@@ -0,0 +1,17 @@
+package main
```
robscott, line 17: Every file with code in it should include a source code header like this: https://github.com/google/new-project#source-code-headers

modeling/data.go
```diff
@@ -0,0 +1,46 @@
...
+	// Outgoing traffic distribution of the same zone
+	OutgoingTraffic map[string]float64
+}
+
+type Stat struct {
+	// Total ratio of traffic that stays in the same zone
+	InZoneTraffic float64
+	// Traffic details for different zones
+	Traffic map[string]*Traffic
+	// Workload balance for different zones -- ratio of incoming traffic / ratio of capacity
```
robscott: Kubernetes standard is to wrap comments at 80 chars.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+func CreateDefaultSim(times uint64) DefaultSimulator {
+	sim := DefaultSimulator{simulationTimes: times}
+	return sim
+}
+
+func (sim DefaultSimulator) Simulate(zones []Zone, slices []Endpointslice) (Stat, error) {
+	// Total number of endpoints and nodes of all zones
+	var totalNodes int
+	var totalPods int
+	// Weighted number of endpoints that in a zone
```
robscott: I think this comment is missing a couple words

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
+package modeling
+
+import (
+	"math/rand"
+	"time"
+)
+
+type DefaultSimulator struct {
```
robscott: I think gofmt requires that all exported types or functions have a comment above them.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+	// Times to run the simulation, larger more accurate but slower
+	simulationTimes uint64
+}
+
+func CreateDefaultSim(times uint64) DefaultSimulator {
+	sim := DefaultSimulator{simulationTimes: times}
+	return sim
```
robscott: can you combine these lines?

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+	// Total number of endpoints and nodes of all zones
+	var totalNodes int
+	var totalPods int
+	// Weighted number of endpoints that in a zone
+	totalWeights := make(map[string]int)
+	// Restruct the slice structure to map structure to use the helper function, zonename - number of pods/nodes
+	zoneNodes := make(map[string]int)
+	zonePods := make(map[string]int)
+	// Composition of the weighted number of endpoints in a zone
+	zoneWeights := make(map[string]map[string]float64)
```
robscott: You're storing a lot of information by zone, could you restructure this into something like `map[string]zoneInfo` where `zoneInfo` contained the number of nodes, pods, and weight info?

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// Calculate the weighted number of pods and composition of different zones
+	for _, zone := range zones {
+		totalWeight := 0.0
+		totalNodes += zone.Nodes
+		totalPods += zone.Endpoints
+		zoneWeights[zone.Name] = make(map[string]float64)
+		for _, slice := range slices {
+			if val, ok := slice.Weights[zone.Name]; ok {
+				tmp := float64(slice.numberOfPods()) * val
```
robscott: It seems like a lot of work to recalculate this for every slice so often, maybe we can store/cache this as part of the data structure?

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+				totalWeight += tmp
+				zoneWeights[zone.Name][slice.Label] = tmp
+			}
+		}
+		totalWeights[zone.Name] = int(totalWeight)
```
robscott: Since even `totalWeights` is indexed by Zone, it likely makes sense to make the value here part of the `zoneInfo` struct I suggested above.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		zoneNodes[zone.Name] = zone.Nodes
+		zonePods[zone.Name] = zone.Endpoints
+	}
+
+	// Total ratio of traffic stays in the same zone
```
robscott: Nit: "that stays"

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+	inZoneTraffic := 0
+	// Detailed traffic information of every single zone
+	detailTraffic := make(map[string]map[string]uint64)
+	// How many requests that hit a zone specifically
+	hitTimes := make(map[string]uint64)
```
robscott: Are both `map[string]uint64` types doing the same thing? Assuming it's a map recording number of requests by zone? It may make sense to actually make that a separate type or something, anytime there's a map of maps it can be a warning sign that code could be refactored a bit. Not quite sure what that looks like yet.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+	for i := uint64(0); i < sim.simulationTimes; i++ {
+		// Simulate the a random request from a zone, incomingZone is the name of the zone making this request
```
robscott: Nit: no need for "the", also should wrap line at 80 chars.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		incomingZone := simulationHelperInt(zoneNodes, totalNodes)
+		// Simulate the slice that request is going to, hitSliceLabel is the lable of the hit slice
+		hitSliceLabel := simulationHelperFloat(zoneWeights[incomingZone], totalWeights[incomingZone])
```
robscott: I think this could be refactored after moving to a new data structure like `zoneInfo`.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		hitSlice := 0
```
robscott: Maybe rename this var so it's clear that it represents the index of a slice?

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		// Get that slice based on the lable
```
robscott: Nit: label

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		for index, slice := range slices {
+			if hitSliceLabel == slice.Label {
+				hitSlice = index
+			}
+		}
```
robscott: Maybe slices should just be a map indexed by name?

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		// Simulate the zone that request is going to, hitZone is the name of the zone receiving this request
+		hitZone := simulationHelperInt(slices[hitSlice].Composition, slices[hitSlice].numberOfPods())
```
robscott: Maybe `destZone` and `destSlice`?

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+
+		// If origin and destination is the same, this is an inzone traffic
```
robscott: Nit: No need for "an"

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		zonePods[zone.Name] = zone.Endpoints
+	}
+
+	// Total ratio of traffic stays in the same zone
+	inZoneTraffic := 0
+	// Detailed traffic information of every single zone
+	detailTraffic := make(map[string]map[string]uint64)
+	// How many requests that hit a zone specifically
+	hitTimes := make(map[string]uint64)
+	for i := uint64(0); i < sim.simulationTimes; i++ {
```
robscott: Maybe the contents of this loop could be moved to a separate function.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		stat.Workload[zone] = stat.Traffic[zone].IncomingTraffic / float64(zonePods[zone]) * float64(totalPods)
+		totalOutgoing := uint64(0)
+		for _, times := range traffic {
+			totalOutgoing += times
+		}
+		stat.Traffic[zone].OutgoingTraffic = make(map[string]float64)
+		for hittenZone, times := range traffic {
+			stat.Traffic[zone].OutgoingTraffic[hittenZone] = float64(times) / float64(totalOutgoing)
+		}
+	}
```
robscott: This could potentially be moved to a separate function.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		// This contributes to the distribution part of the result
+		detailTraffic[incomingZone][hitZone]++
+	}
+
+	var stat Stat
+	stat.InZoneTraffic = float64(inZoneTraffic) / float64(sim.simulationTimes)
+	stat.Traffic = make(map[string]*Traffic)
+	stat.Workload = make(map[string]float64)
+	// Calculte the metrics of the result, should be quite straightforward
```
robscott: Nit: Calculate

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+			detailTraffic[incomingZone] = make(map[string]uint64)
+		}
+		// This contributes to the distribution part of the result
+		detailTraffic[incomingZone][hitZone]++
+	}
+
+	var stat Stat
+	stat.InZoneTraffic = float64(inZoneTraffic) / float64(sim.simulationTimes)
+	stat.Traffic = make(map[string]*Traffic)
+	stat.Workload = make(map[string]float64)
```
robscott: This might be easier to work with if a generic stat data structure containing both traffic and workload information was indexed by zone, something like `map[string]stat` where string represented the zone. I think that would end up having more significant consequences though and may not be worth the effort.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+	// Calculte the metrics of the result, should be quite straightforward
+	for zone, traffic := range detailTraffic {
+		stat.Traffic[zone] = new(Traffic)
+		stat.Traffic[zone].OutgoingZone = zone
```
robscott: `OutgoingZone` seems confusing here since it does not seem to be tied to `OutgoingTraffic`.

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+		stat.Traffic[zone].IncomingTraffic = float64(hitTimes[zone]) / float64(sim.simulationTimes)
+		stat.Workload[zone] = stat.Traffic[zone].IncomingTraffic / float64(zonePods[zone]) * float64(totalPods)
+		totalOutgoing := uint64(0)
+		for _, times := range traffic {
+			totalOutgoing += times
+		}
+		stat.Traffic[zone].OutgoingTraffic = make(map[string]float64)
+		for hittenZone, times := range traffic {
```
robscott: Maybe `destZone`

modeling/default-simulator.go
```diff
@@ -0,0 +1,132 @@
...
+	random := rand.New(rand.NewSource(time.Now().UnixNano()))
+	tmp := random.Intn(total)
+	acc := 0
+	for key, val := range probs {
+		upper := acc + val
+		if tmp >= acc && tmp < upper {
+			return key
+		}
+		acc = upper
+	}
```
robscott: At first glance it's not obvious to me what these variables represent or why they are named what they are, specifically `tmp` and `acc`. Some comments explaining the logic and/or renaming some of the vars might help.

modeling/interface.go
```diff
@@ -0,0 +1,15 @@
+package modeling
+
+type RoutingAlgorithm interface {
+	// This interface is to create endpointslices based on the current zones and the rouing algorithm
+	//		Input: zones that involved in the routing
+	//		Output: endpointslices that created based on the routing rules
+	CreatingSlices([]Zone) ([]Endpointslice, error)
```
robscott: Maybe `CreateSlices`?

modeling/model.go
```diff
@@ -0,0 +1,122 @@
+package modeling
+
+import (
+	"errors"
+	"fmt"
+)
+
+// slicecapacity is the number of max endpoints per slice
```
robscott: This comment should be above the field name, not the struct. Although I think gofmt will also require a separate comment above the struct since it's being exported.

modeling/model.go
```diff
@@ -0,0 +1,122 @@
...
+	zones         []Zone
+	slices        []Endpointslice
+	sliceCapacity uint
+	lastResult    Stat
+	alg           RoutingAlgorithm
+	simulator     TrafficSimulator
+}
+
+// Create a model with zones, routing algorithm and traffic simulator
+//		Based on the algorithm, create the endpointslices
```
robscott: No need for second line of comments to be indented this far.

modeling/model.go
```diff
@@ -0,0 +1,122 @@
...
+		zones:         zones,
+		sliceCapacity: 100,
+		alg:           alg,
+		simulator:     sim,
+	}
+	if slices, err := model.alg.CreatingSlices(model.zones); err != nil {
+		panic(err)
+	} else {
+		model.slices = slices
+	}
```
robscott:
```suggestion
	slices, err := model.alg.CreatingSlices(model.zones)
	if err != nil {
		return nil, err
	} 
	model.slices = slices
```

modeling/model.go
```diff
@@ -0,0 +1,122 @@
...
+	return model, nil
+}
+
+func (m *Model) SetSliceCapacity(newCapacity uint) {
+	m.sliceCapacity = newCapacity
+}
```
robscott: I think I'd remove this since I don't think it's actually being used. It may make sense to make this a param for the NewModel function though.

modeling/model.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Start simulation based on the zones and endpointslices
+func (m *Model) StartSimulation() error {
+	if stat, err := m.simulator.Simulate(m.zones, m.slices); err != nil {
+		panic(err)
+	} else {
+		m.lastResult = stat
+	}
```
robscott:
```suggestion
	stat, err := m.simulator.Simulate(m.zones, m.slices)
	if err != nil {
		return err
	}
	m.lastResult = stat
```

modeling/model.go
```diff
@@ -0,0 +1,122 @@
...
+func (m *Model) UpdateAlgorithm(alg RoutingAlgorithm) error {
+	if alg == nil {
+		return errors.New("Empty algorithm")
+	}
+	m.alg = alg
+	if slices, err := m.alg.CreatingSlices(m.zones); err != nil {
+		panic(err)
+	} else {
+		m.slices = slices
+	}
```
robscott: Same suggestion around error handling here.

modeling/model.go
```diff
@@ -0,0 +1,122 @@
...
+		panic(err)
+	} else {
+		m.lastResult = stat
+	}
+	return nil
+}
+
+// Update the algorithm
+//		create the new endpointslices based on the new algorithm
+func (m *Model) UpdateAlgorithm(alg RoutingAlgorithm) error {
```
robscott: It looks like this is also not currently used, do you think it will be useful in the future?

modeling/model.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	return nil
+}
+
+func (m *Model) PrintLastResult() error {
+	var totalPods int
+	for _, zone := range m.zones {
+		totalPods += zone.Endpoints
+	}
+	_, numberOfSlices, _ := m.GetEndpointslices()
```
robscott: Always handle errors. For the first return value, it looks like we're not using it and this is the only place it's being called, so it may be worth removing it.

modeling/model.go
```diff
@@ -0,0 +1,122 @@
...
+		// Suppose to be initialized outside, if empty panic!
+		panic(errors.New("Can't get empty zones"))
+	}
+	return m.zones, nil
+}
+
+func (m *Model) GetEndpointslices() ([]Endpointslice, uint, error) {
+	if len(m.slices) == 0 {
+		return nil, 0, errors.New("Can't get empty slices")
+	}
```
robscott: Is this necessarily an error? Could you just return 0/empty values?

modeling/model.go
```diff
@@ -0,0 +1,122 @@
...
+	totalSlice := uint(0)
+	for _, slice := range m.slices {
+		pods := slice.numberOfPods()
+		totalSlice += uint(pods) / m.sliceCapacity
+		if uint(pods)%m.sliceCapacity != 0 {
+			totalSlice++
+		}
+	}
```
robscott: This is confusing to me, wouldn't `totalSlice` represent `len(m.slices`)?

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"errors"
+	"math"
+)
+
+type SharedGlobalAlgorithm struct {
+	// Weight of global endpointslces
+	globalWeight float64
+	// Threshold of global endpointslces that if the total number of endpoints <= threshold, all endpoints go to global endpointslice
+	//		Int should be enough
```
robscott: Nit: EndpointSlices, wrap to 80 chars

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,82 @@
...
+	globalTheshold int
+}
+
+func (alg SharedGlobalAlgorithm) CreatingSlices(zones []Zone) ([]Endpointslice, error) {
+	// Total number of endpoints and nodes of all zones
+	var totalPods, totalNodes int
+	// This assumes the incoming traffic of a zone is proportional to the number of nodes of that zone
+	tafficRatio := make([]float64, len(zones))
```
robscott: Nit: trafficRatio

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,82 @@
...
+	"errors"
+	"math"
+)
+
+type SharedGlobalAlgorithm struct {
+	// Weight of global endpointslces
+	globalWeight float64
+	// Threshold of global endpointslces that if the total number of endpoints <= threshold, all endpoints go to global endpointslice
+	//		Int should be enough
+	globalTheshold int
```
robscott: Nit: globalThreshold

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,82 @@
...
+	var endpointslices []Endpointslice
+	// The 'big' global slice -- might be split into many small global slices when the number of endpoints > required number of endpoints/endpoinslice, i.e. 100 for default
+	//		not able to divide the big one into smaller ones that the contributions may vary and there is no need to do so either.
+	var globalSlice Endpointslice
+	globalSlice.Label = "global"
+	globalSlice.Composition = make(map[string]int)
+	globalSlice.Weights = make(map[string]float64)
+	for index, zone := range zones {
+		var globalPods int
+		// If total pods <= threshold, all pods go to global slice
```
robscott, line 67: I think we may want to rethink this part of the algorithm or at least make the threshold smaller.

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,82 @@
...
+		if totalPods <= alg.globalTheshold {
+			globalPods = zone.Endpoints
+		} else {
+			// Otherwise calculte the global contribution of current zone based on the global weight and the deviation of this zone
```
robscott: Nit: calculate

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,82 @@
...
+			globalPods = int(math.Min(math.Max(0.0, deviation[index])/alg.globalWeight, float64(zone.Endpoints)))
+		}
+
+		globalSlice.Composition[zone.Name] = globalPods
+		globalSlice.Weights[zone.Name] = alg.globalWeight
+
+		// Calculate how many endpoints remain in the local zone
+		var slice Endpointslice
+		slice.Label = zone.Name
```
robscott: To really test this properly we'll want to get an idea of the number of EndpointSlices involved. As I understand it right now, there is 1 global EndpointSlice + 1 for each zone. That's likely enough for simulating data, but we'll need to be sure that our calculations can translate that into the number of EndpointSlices that will actually be required to represent this data. 

If we want to keep the current approach of this type representing endpoints allocated to a zone, we should probably rename this from EndpointSlice so it's not confused with that data structure. Maybe something like `ZonalEndpoints`?

robscott (commented):
Thanks for all your work on this, it's a great start! I've added a lot of comments here but most of them are relatively minor nitpicky things.

modeling/shared-global-algorithm.go
ckyiricky: refactored endpointslice to endpointslicegroup

modeling/shared-global-algorithm.go
ckyiricky: fixed

modeling/shared-global-algorithm.go
ckyiricky: We can think about how to update this algorithm as well as some other algorithms in future versions

modeling/shared-global-algorithm.go
ckyiricky: fixed

modeling/shared-global-algorithm.go
ckyiricky: fixed

modeling/shared-global-algorithm.go
ckyiricky: fixed

modeling/model.go
ckyiricky: m.slice is groups of endpointslices (refactored data structure) that it could contain multiple endpointslices in one group that share a same label. len(m.slice) can differ from the totalSlice

modeling/model.go
ckyiricky: fixed

modeling/model.go
ckyiricky: fixed, the first returned value is the endpointslicegroups, I think its better to keep it private inside the package since this filed is computed based on zones and algorithms. This method runs as a getter for the endpointslicegroups.

modeling/model.go
ckyiricky: I think this might be used in future that we run different algorithms, so this method has been reserved

modeling/model.go
ckyiricky: fixed

modeling/model.go
ckyiricky: fixed

modeling/model.go
ckyiricky: removed and change the capacity field to be public, the NewModel keeps the default value as 100, users has to manually change that if they want another value (no default value for parameter in go :( )

modeling/model.go
ckyiricky: fixed

modeling/model.go
ckyiricky: hopefully it has been fixed, not sure how the tab in gofmt will be looked like on github

modeling/model.go
ckyiricky: fixed

modeling/interface.go
ckyiricky: fixed and refactored

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: Maybe update this structure in future versions along with the input/output update version

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: default-simulator has been removed

modeling/default-simulator.go
ckyiricky: default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: refactored but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/default-simulator.go
ckyiricky: fixed but default-simulator has been removed

modeling/data.go
ckyiricky: fixed

main.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed hope its clearer now

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed and renamed

main.go
ckyiricky: fixed

main.go
ckyiricky: Gonna update in other PRs

main.go
ckyiricky: fixed

main.go
ckyiricky: Gonna update this part in other PRs

ckyiricky:
Updates:
    fix several typos and format problems
    add wrapper struct for zones
    refactored the RoutingAlgorithm and TrafficSimulator inputs
    remove default-simulator
    add theoretical-simulator instead which calculated the theoretical probability of traffic distribution

modeling/data.go
```diff
@@ -0,0 +1,106 @@
...
+	// Total number of endpoints of all zones
+	totalEndpoints int
+	// Detailed information of each zone, zone name - exact zone
+	zoneDetails map[string]Zone
+}
+
+// EndpointSliceGroup represents all the endpointslices under a same label, one
+// group may be made up by many endpointslices (when the number of endpoints
+// excceeds the capacity of one endpointslice). Since for now there is no need
+// to know how the group is composed, we keep them as a whole
```
robscott: Nit: K8s objects are generally referred to with pascal case, so in this case EndpointSlices.

modeling/data.go
```diff
@@ -0,0 +1,106 @@
...
+type EndpointSliceGroup struct {
+	// Label of this endpointslice
```
robscott: Nit: slice -> group, same for fields below

robscott: Maybe add some more info about how this will be used. Something like "Label is a unique identifier for an EndpointSliceGroup. This often represents a topology label such as a zone name that the group will be consumed from."

modeling/data.go
```diff
@@ -0,0 +1,106 @@
...
+limitations under the License.
+*/
+
+package modeling
+
+import "errors"
+
+// Zone mock struct for real zones in k8s
+type Zone struct {
+	// Number of nodes of this zone
```
robscott: As hard as it can be to get used to, it's ideal for every comment above a field to start with the name of the field.

> Inside a package, any comment immediately preceding a top-level declaration serves as a doc comment for that declaration. Every exported (capitalized) name in a program should have a doc comment.

> Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared.

> If every doc comment begins with the name of the item it describes, you can use the doc subcommand of the go tool and run the output through grep.

https://golang.org/doc/effective_go.html#commentary

modeling/data.go
```diff
@@ -0,0 +1,106 @@
...
+	Nodes int
+	// Number of endpoitns of this zone
```
robscott: Nit: endpoints in

modeling/data.go
```diff
@@ -0,0 +1,106 @@
...
+	IncomingTraffic float64
+	// Outgoing traffic distribution of the same zone
+	OutgoingTraffic map[string]float64
+}
+
+// Stat is to collect metrics of a simulation result
+type Stat struct {
+	// Total ratio of traffic that stays in the same zone
+	InZoneTraffic float64
+	// Traffic details for different zones, zone name - traffic details
```
robscott: Maybe `TrafficDetail groups Traffic by zone name`.

modeling/data.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Helper function to calculate number of endpoints of a specific endpointslice
+func (e EndpointSliceGroup) numberOfPods() int {
+	total := 0
+	for _, pods := range e.Composition {
+		total += pods
+	}
+	return total
+}
```
robscott: Can we represent this with a field on EndpointSliceGroup instead?

modeling/data.go
```diff
@@ -0,0 +1,106 @@
...
+
+func createZoneinfos(zones []Zone) (zoneInfos, error) {
```
robscott: Nit: createZoneInfos

modeling/data.go
```diff
@@ -0,0 +1,106 @@
...
+	Endpoints int
+	// Name of this zone
+	Name string
+	// Number of pods / total number of pods of all zones
+	endpointsRatio float64
+	// Number of nodes / total number of nodes of all zones
+	nodesRatio float64
+}
+
+type zoneInfos struct {
```
robscott: Maybe this should be `clusterInfo` since it represents the full state of a cluster (at least for the scope of this).

modeling/model.go
```diff
@@ -0,0 +1,130 @@
...
+	fmt.Println("----------------------------------------------")
+
+	for zone, workload := range lastResult.Workload {
+		fmt.Printf("Workload for %s : \t %.2f%% \n", zone, workload*100)
+	}
+	return nil
+}
+
+// GetEndpointslices gets the created endpoitnslicegroups, number of
```
robscott: Nit: EndpointSliceGroups

modeling/model.go
```diff
@@ -0,0 +1,130 @@
...
+// endpointslices
+func (m *Model) GetEndpointslices() (map[string]EndpointSliceGroup, uint, error) {
```
robscott: Nit: GetEndpointSliceGroups().

robscott: Do we need to return the number of EndpointSlices here? If that's a useful number, it might be better to move it out to a separate function.

modeling/interface.go
```diff
@@ -0,0 +1,34 @@
...
+
+package modeling
+
+// RoutingAlgorithm interface for different routing algorithms
+type RoutingAlgorithm interface {
+	//This interface is to create endpointslices based on the current zones and
+	//the rouing algorithm
+	//	Input: zones that involved in the routing
+	//	Output: endpointslices that created based on the routing rules
+	CreateSlices(zoneInfos) (map[string]EndpointSliceGroup, error)
```
robscott: Maybe something like: This interface translates clusterInfo into EndpointSliceGroups.

robscott: Should likely rename function to `CreateSliceGroups`

modeling/interface.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// TrafficSimulator interface for different simulators
+type TrafficSimulator interface {
+	//This interface is to simulate the traffic among the zones
+	//	Input: zones and endpointslices
+	//	Output: Simulation results
+	Simulate(zoneInfos, map[string]EndpointSliceGroup) (Stat, error)
```
robscott: Maybe: This interface simulates traffic with the provided clusterInfo and EndpointSliceGroups and returns a SimulationResult.

modeling/data.go
```diff
@@ -0,0 +1,106 @@
...
+	// Name of a specific zone
+	ZoneName string
+	// Traffic that the same zone received
+	IncomingTraffic float64
+	// Outgoing traffic distribution of the same zone
+	OutgoingTraffic map[string]float64
+}
+
+// Stat is to collect metrics of a simulation result
+type Stat struct {
```
robscott: Maybe we should just call this SimulationResult

modeling/model.go
robscott: In general we avoid adding any public functions like this unless they're actually going to be immediately used since once it exists it's impossible to safely remove. I'd lean towards removing this function until it's going to be used.

modeling/model.go
```diff
@@ -0,0 +1,130 @@
...
+	if len(m.slices) == 0 {
+		// To avoid the situation m.slices is not nil but is empty, intentionally
+		// return the nil for the first value
+		return nil, 0, nil
+	}
+	totalSlice := uint(0)
```
robscott: Nit: totalSlices

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,102 @@
...
+
+// SharedGlobalAlgorithm takes multiple zones as input and output endpointslices
+// composition based on their nodes and endpoints:
+// 1. Endpointslices will be considered global by default
+// 2. Once the number of endpoints in a zone reaches a specified threshold,
+//	  zone-specific EndpointSlices will begin to be produced.
+// 3. Kube Proxy will consume both global and zone specific EndpointSlices.
+//	  Endpoints in the same zone will be given a higher weight.
+type SharedGlobalAlgorithm struct {
+	// Weight of global endpointslices
```
robscott: This likely needs more explanation.

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,102 @@
...
+	globalWeight float64
+	// Threshold of global endpointslices that if the total number of endpoints
+	// <= threshold, all endpoints go to global endpointslice
+	//	 Int should be enough
```
robscott: I'd remove this last line here

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,102 @@
...
+	deviation := make(map[string]float64)
+
+	for _, zone := range zones.zoneDetails {
+		// Calculate the deviation based on the capacity(endpoints) and
+		// traffic(nodes) ratio
+		deviation[zone.Name] = float64(zone.Endpoints) - float64(zones.totalEndpoints)*zone.nodesRatio
+	}
+
+	// Output endpointslices
+	endpointslices := make(map[string]EndpointSliceGroup)
```
robscott: endpointslices -> sliceGroups

robscott (commented):
Sorry this review took so long! I think this is getting close to ready, mostly just have some nits around naming.

modeling/data.go
ckyiricky: I think it might not be perfectly accurate to refer all the zones as a cluster. Can we just keep them as zones since cluster might bring in a new 'topology' relation.

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: modified the comment above Zone.

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: pending based on future usage, currently remain the same since only one place calls the method.

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: renamed to regionInfo

modeling/model.go
ckyiricky: fixed

modeling/model.go
ckyiricky: fixed, split the function into two

modeling/interface.go
ckyiricky: refactored

modeling/interface.go
ckyiricky: fixed

modeling/data.go
ckyiricky: refactored

modeling/model.go
ckyiricky: fixed

modeling/shared-global-algorithm.go
ckyiricky: Will add more detailed explanation afterwards.

modeling/shared-global-algorithm.go
ckyiricky: fixed

modeling/shared-global-algorithm.go
ckyiricky: fixed

ckyiricky:
Updates:
Refactor zoneInfo to regionInfo
Remove unnecessary method: UpdateAlgorithm
Fix some typos

modeling/theoretical-simulator.go
```diff
@@ -0,0 +1,69 @@
...
+		zoneEndpoints[zoneName] = map[string]float64{}
+		for destZoneName, endpoints := range endpointsDistribution {
+			zoneEndpoints[zoneName][destZoneName] = endpoints / zoneTotal
+		}
+	}
+
+	// calculate result stat
+	var stat SimulationResult
+	stat.TrafficDetail = map[string]Traffic{}
+	stat.Workload = map[string]float64{}
```
robscott: Nit: replace stat with more representative name, maybe `sr` or `simResult`?

modeling/data.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Traffic represents the detailed traffic infomation of a zone
+type Traffic struct {
+	// Name of a specific zone
+	ZoneName string
+	// Traffic that the same zone received
+	IncomingTraffic float64
+	// Outgoing traffic distribution of the same zone
+	OutgoingTraffic map[string]float64
```
robscott: Nit: These names are repetitive. Could likely just use `Incoming` and `Outgoing` to end up with `traffic.Incoming` instead of `traffic.IncomingTraffic`.

modeling/data.go
```diff
@@ -0,0 +1,108 @@
...
+
+// EndpointSliceGroup represents all the EndpointSlices under a same label, one
+// group may be made up by many EndpointSlices (when the number of endpoints
+// excceeds the capacity of one EndpointSlice). Since for now there is no need
+// to know how the group is composed, we keep them as a whole
+type EndpointSliceGroup struct {
+	// Label is a unique identifier for an EndpointSliceGroup. This often
+	// represents a topology label that the group will be consumed from.
+	Label string
+	// Contribution of endpoints in this group from different zones
```
robscott: Nit: comment should start with field name, at least on public fields.

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,100 @@
...
+		deviation[zone.Name] = float64(zone.Endpoints) - float64(region.totalEndpoints)*zone.nodesRatio
+	}
+
+	// Output EndpointSlices
+	sliceGroups := make(map[string]EndpointSliceGroup)
+	// The 'big' global slice -- might be split into many small global slices
+	// when the number of endpoints > required number of endpoints/endpoinslice,
+	// i.e. 100 for default not able to divide the big one into smaller ones that
+	// the contributions may vary and there is no need to do so either.
+	var globalSlice EndpointSliceGroup
```
robscott: Nit: update naming to include "group", maybe `globalSG` or `globalSliceGroup`

modeling/shared-global-algorithm.go
```diff
@@ -0,0 +1,100 @@
...
+			// Otherwise calculate the global contribution of current zone based
+			// on the global weight and the deviation of this zone
+			globalEndpoints = int(math.Min(math.Max(0.0, deviation[name])/alg.globalWeight, float64(zone.Endpoints)))
+		}
+
+		globalSlice.Composition[name] = globalEndpoints
+		globalSlice.Weights[name] = alg.globalWeight
+
+		// Calculate how many endpoints remain in the local zone
+		var slice EndpointSliceGroup
```
robscott: Nit: `sg` or `sliceGroup`

modeling/model.go
```diff
@@ -0,0 +1,111 @@
...
+
+// Model wrapper class for the simulation components
+type Model struct {
+	slices    map[string]EndpointSliceGroup
+	alg       RoutingAlgorithm
+	simulator TrafficSimulator
+
+	Region  regionInfo
+	Results []SimulationResult
+	// number of max endpoints per slice
```
robscott: Nit: Comment should start with field name.

robscott (commented):
Thanks for the update! I think this is my last and smallest round of nits, hopefully we can get this in today.

robscott:
One more note, please squash all these commits together as part of the next update to get this PR ready to merge.

modeling/model.go
ckyiricky: fixed

modeling/shared-global-algorithm.go
ckyiricky: fixed

modeling/shared-global-algorithm.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/theoretical-simulator.go
ckyiricky: fixed

ckyiricky:
Update:
Fix some comment format 
Rename some variables with names more related to refactored types
Fix some minor bugs and convention issues
Squash all commits into one

robscott (approved):
Thanks! Looks good to me.

PR #6, opened 2020-07-24
Implement routing algorithm 2

ckyiricky:
Implement algorithm 2 :
    EndpointSlices with unique endpoints for each zone

### DIFF
--- a/modeling/local-slice-algorithm.go
+++ b/modeling/local-slice-algorithm.go
@@ -1,0 +1,154 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package modeling
+
+import (
+	"fmt"
+)
+
+// LocalSliceAlgorithm 'borrows' and 'rents' endpoints from other zones to make
+// the local EndpointSlice balanced with the incoming traffic (number of nodes
+// distribution):
+// 1. The proportion of incoming traffic for a zone is calculated based on the
+// proportion of nodes or cores in the zone.
+// 2. This is compared with the proportion of endpoints in the zone to calculate
+// the deviation from the expected number of endpoints in a perfectly balanced
+// system.
+// 3. EndpointSlices in zones with less endpoints than expected will receive
+// endpoints from zones that have more endpoints than expected.
+type LocalSliceAlgorithm struct{}
+
+// endpointDeviation stores the deviation between expected number of endpoints
+// and actual number of endpoints of a zone
+type endpointDeviation struct {
+	name string
+	// deviation is the absolute value of the deviation = actual - expected
+	deviation int
+}
+
+// endpointsList maintains the order of the endpoints available/needed, used as
+// a list
+type endpointsList struct {
+	byZone []endpointDeviation
+}
+
+// push to back
+func (el *endpointsList) push(newValue endpointDeviation) {
+	el.byZone = append(el.byZone, newValue)
+}
+
+// pop the front
+func (el *endpointsList) pop() {
+	if len(el.byZone) == 0 {
+		return
+	}
+	el.byZone = el.byZone[1:]
+}
+
+// CreateSliceGroups creates sliceGroups with 'one local EndpointSliceGroup per
+// zone' policy
+func (alg LocalSliceAlgorithm) CreateSliceGroups(region regionInfo) (map[string]EndpointSliceGroup, error) {
+	if region.zoneDetails == nil {
+		return nil, fmt.Errorf("zoneDetail should not be nil")
+	}
+	sliceGroups := map[string]EndpointSliceGroup{}
+	// endpointsAvailable stores zones with number of endpoints available
+	endpointsAvailable := endpointsList{}
+	// endpointsNeeded stores zones with number of endpoints needed
+	endpointsNeeded := endpointsList{}
+	for zoneName, zone := range region.zoneDetails {
+		var localGroup EndpointSliceGroup
+		localGroup.Label = zoneName
+		// this local sliceGroup should only receive traffic from current zone,
+		// this map tracks weights of traffic from different zones to this
+		// sliceGroup
+		localGroup.Weights = map[string]float64{zoneName: 1.0}
+		// this map stores composition of this sliceGroup, zoneName - number of
+		// endpoints from zoneName
+		localGroup.Composition = map[string]int{}
+
+		// calculate expected number of endpoints based on the proportion of
+		// nodes in this zone
+		expectedEndpoints := int(zone.nodesRatio * float64(region.totalEndpoints))
+		// deviation: a negative value means need more endpoints from other
+		// sliceGroups, a positive value means need give out endpoints to other
+		// sliceGroups
+		deviation := zone.Endpoints - expectedEndpoints
+		if deviation > 0 {
+			endpointsAvailable.push(endpointDeviation{name: zoneName, deviation: deviation})
+			localGroup.Composition[zoneName] = int(expectedEndpoints)
+		} else {
+			endpointsNeeded.push(endpointDeviation{name: zoneName, deviation: -deviation})
+			localGroup.Composition[zoneName] = zone.Endpoints
+		}
+
+		sliceGroups[zoneName] = localGroup
+	}
+
+	err := balanceSliceGroups(&endpointsAvailable, &endpointsNeeded, sliceGroups)
+	if err != nil {
+		return nil, err
+	}
+	return sliceGroups, nil
+}
+
+// balanceSliceGroups distributes endpoints from zones with extra endpoints to
+// EndpointSliceGroups for zones with insufficient endpoints.
+func balanceSliceGroups(endpointsAvailable *endpointsList, endpointsNeeded *endpointsList, sliceGroups map[string]EndpointSliceGroup) error {
+	for _, receiveZone := range endpointsNeeded.byZone {
+		// the available list is empty while there are still endpoints in
+		// need. This can happen when the approximation on deviation
+		// (calculated above) ends up in asymmetric sums of endpoints
+		// available and endpoints in need (in need > available)
+		if len(endpointsAvailable.byZone) == 0 {
+			// in this case, we do nothing, ignore the extra endpoints needed.
+			return nil
+		}
+		for index := 0; index < len(endpointsAvailable.byZone); {
+			sendZone := endpointsAvailable.byZone[index]
+			if sendZone.deviation == receiveZone.deviation {
+				sliceGroups[receiveZone.name].Composition[sendZone.name] = sendZone.deviation
+				endpointsAvailable.pop()
+				break
+			}
+			if sendZone.deviation > receiveZone.deviation {
+				sliceGroups[receiveZone.name].Composition[sendZone.name] = receiveZone.deviation
+				endpointsAvailable.byZone[index].deviation -= receiveZone.deviation
+				break
+			}
+			if sendZone.deviation < receiveZone.deviation {
+				sliceGroups[receiveZone.name].Composition[sendZone.name] = sendZone.deviation
+				receiveZone.deviation -= sendZone.deviation
+				endpointsAvailable.pop()
+				continue
+			}
+		}
+		endpointsNeeded.pop()
+	}
+	// all endpoints should have been distributed properly. This happens when
+	// the sum of approximated available endpoints > sum of approximated
+	// endpoints in need
+	if len(endpointsAvailable.byZone) != 0 {
+		// in this case, we assign those extra endpoints to its local
+		// endpointSliceGroups
+		for _, extraEndpoints := range endpointsAvailable.byZone {
+			sliceGroups[extraEndpoints.name].Composition[extraEndpoints.name] += extraEndpoints.deviation
+			endpointsAvailable.pop()
+		}
+	}
+	return nil
+}


ckyiricky:
The action check might have some cache issues, the local check with the errcheck tool has been passed.

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"container/list"
+	"fmt"
+	"math"
+)
+
+// LocalSliceAlgorithm 'borrows' and 'rents' endpoints from other zones to make
+// the local EndpointSlice balanced with the incoming traffic (number of nodes
+// distribution)
```
robscott: This would be a good place to put a more detailed explanation of the algorithm. IE, how does it decide to borrow endpoints and from where?

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,121 @@
...
+type endpointDeviation struct {
+	name      string
+	deviation float64
+}
+
+// CreateSliceGroups creates sliceGroups with 'one local EndpointSliceGroup per
+// zone' policy
+func (alg LocalSliceAlgorithm) CreateSliceGroups(region regionInfo) (map[string]EndpointSliceGroup, error) {
+	if region.zoneDetails == nil {
+		return nil, fmt.Errorf("can't create EndpointSlices with 0 number of zone")
```
robscott: Maybe `zoneDetails can't be nil` or `expected zoneDetails to not be nil`.

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	sliceGroups := map[string]EndpointSliceGroup{}
+	// stack stores 'endpointDeviation' -- the info of how should number of
+	// endpoints in a zone be adjusted to match the ratio balance
```
robscott: Nit: stack -> endpointStack

Maybe: `the value showing how endpoints in a zone should be adjusted to match the ratio balance`

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,121 @@
...
+	endpointStack := list.New()
+	for name, zone := range region.zoneDetails {
+		var localGroup EndpointSliceGroup
+		localGroup.Label = name
+		localGroup.Weights = map[string]float64{name: 1.0}
```
robscott: Maybe this could use a comment to better explain what's happening. On first glance it looks like it's initializing something where name is "1.0". Maybe explain that this map is used to track weights by zone name?

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,121 @@
...
+		localGroup.Composition = map[string]int{}
```
robscott, line 82: What do the string and int represent here? Assuming it's zone name and number of endpoints? Maybe add a comment to explain.

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,121 @@
...
+
+		deviation := math.Round(zone.nodesRatio*float64(region.totalEndpoints) - float64(zone.Endpoints))
```
robscott: Also could use a comment here to explain the math here.

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,121 @@
...
+		// assign the endpoints to its local sliceGroup
+		localGroup.Composition[name] = int(math.Min(float64(zone.Endpoints), float64(zone.Endpoints)+deviation))
+		if deviation == 0 {
+			continue
+		}
+		sliceGroups[name] = localGroup
+		err := createGroupHelper(endpointStack, endpointDeviation{name: name, deviation: deviation}, sliceGroups)
```
robscott: It looks like this function contains the vast majority of the logic here but that's not entirely clear from the name, maybe something like `balanceGroups` would make sense here?

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,121 @@
...
+	// all endpoints should have been distributed properly
+	if endpointStack.Len() != 0 {
+		return nil, fmt.Errorf("stack of endpoints should be empty but with size %d", endpointStack.Len())
+	}
+	return sliceGroups, nil
+}
+
+// createGroupHelper helps to 'borrow' or 'rent' endpoints to other
+// EndpointSliceGroups
+func createGroupHelper(stack *list.List, receiver endpointDeviation, sliceGroups map[string]EndpointSliceGroup) error {
```
robscott: I may be wrong here, but this seems like it could be simplified. I think all we actually need to do for this entire algorithm is calculate the expected endpoints for each zone given the proportion of nodes in the zone and calculating the same proportion of endpoints. Then you can simply loop through zones once to get all the endpoints available to give from each zone, and then distribute them to zones that need them in a second loop. Not sure if that makes sense here, but I think it could end up being simpler to implement and understand.

Example pseudocode:

```
endpointsAvailable := map[string]int{}
for zone in zones {
  zone.deviation := zone.endpoints - expectedEndpoints(zone)
  if zone.deviation > 0 {
    endpointsAvailable[zone.name] := zone.deviation
  }
}

for zone in zones {
  if zone.deviation < 0 {
    takeEndpoints(zone, endpointsAvailable)
  }
}
```

I know it's more complicated than that pseudocode, but maybe that general approach would be simpler?

robscott (commented):
Thanks for the work on this, exciting to see another algorithm in practice!

modeling/local-slice-algorithm.go
ckyiricky: Agreed, this seems to be a neater implementation and much more straightforward compared to the previous one which is quite intricate to understand at the first glance.

ckyiricky: Real implementation looks a little bit different from the pseudo code above, but the idea remains the same

modeling/local-slice-algorithm.go
ckyiricky: renamed to balanceSliceGroups

modeling/local-slice-algorithm.go
ckyiricky: done

modeling/local-slice-algorithm.go
ckyiricky: done

modeling/local-slice-algorithm.go
ckyiricky: done

modeling/local-slice-algorithm.go
ckyiricky: fixed

modeling/local-slice-algorithm.go
ckyiricky: fixed

modeling/local-slice-algorithm.go
ckyiricky: added more explanation of the algorithm

ckyiricky:
[Update]
Fix some typos
Add more comments
Change the logic of balancing sliceGroups

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, err
+	}
+	return sliceGroups, nil
+}
+
+// balanceSliceGroups helps to 'rent' endpoints from zones have extra endpoints
+// to EndpointSliceGroups that have insufficient ones
+func balanceSliceGroups(endpointsAvailable *list.List, endpointsNeeded map[string]int, sliceGroups map[string]EndpointSliceGroup) error {
+	for receiveZone, deviation := range endpointsNeeded {
+		for {
```
robscott: This feels dangerous - can we guarantee that this loop will always exit?

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,132 @@
...
+			// the available stack is empty while there are still endpoints in
+			// need causes an error. This can happen when the approximation on
+			// deviation (calculated above) ends up in asymmetric sums of
+			// endpoints available and endpoints in need (in need > available)
+			if endpointsAvailable.Back() == nil {
+				return fmt.Errorf("endpoints available are less than needed")
+			}
```
robscott: I don't think we want to have an error like this with valid input. I think it would be better to just choose the least bad placement for the extra endpoint or if there's no endpoints available like here not error out. Longer term I think we should update this approach to include some kind of weighting or something.

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,132 @@
...
+			sendZone, ok := endpointsAvailable.Back().Value.(endpointDeviation)
+			endpointsAvailable.Remove(endpointsAvailable.Back())
```
robscott: Why not wait to remove until after we know we're going to use all these endpoints?

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,132 @@
...
+			if sendZone.deviation == deviation {
+				sliceGroups[receiveZone].Composition[sendZone.name] = deviation
+				break
+			}
+			if sendZone.deviation > deviation {
+				// there are extra endpoints remaining, put them back in the
+				// available stack, wait for the next customer
+				sliceGroups[receiveZone].Composition[sendZone.name] = deviation
+				sendZone.deviation -= deviation
+				endpointsAvailable.PushBack(sendZone)
```
robscott: Related to comment above, I don't think we should remove and push back.

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,132 @@
...
+	err := balanceSliceGroups(endpointsAvailable, endpointsNeeded, sliceGroups)
+	if err != nil {
+		return nil, err
+	}
+	return sliceGroups, nil
+}
+
+// balanceSliceGroups helps to 'rent' endpoints from zones have extra endpoints
+// to EndpointSliceGroups that have insufficient ones
+func balanceSliceGroups(endpointsAvailable *list.List, endpointsNeeded map[string]int, sliceGroups map[string]EndpointSliceGroup) error {
```
robscott: This function still feels pretty complex. Maybe it would be possible to restructure this a bit further? I think the balancing could benefit from some kind of struct like this. This is pretty rough and probably doesn't work as is, but hopefully it provides an idea of how it could be structured. Something like this would be a great place to add some unit tests.

```golang
type endpointsAvailable struct {
  byZone map[string]int
}

func (ea *endpointsAvailable) takeEndpoints(n int) map[string]int {
  takenByZone := map[string]int{}
  for zone, num := range ea.byZone {
    if num >= n {
      takenByZone[zone] = n
      ea.byZone[zone] = num - n
      break
    } else {
      takenByZone[zone] = num
      ea.byZone[zone] = 0
      n -= num
    }
  }
  return takenByZone
}
```

robscott (commented):
Thanks for the updates here! I do think this is looking better, just had some more ideas on how we could further simplify this.

modeling/local-slice-algorithm.go
ckyiricky: Gonna change this list structure to map instead, I don't think its allowed to modify the value in the list directly. So I pop it out everytime.

modeling/local-slice-algorithm.go
ckyiricky: I don't think it would go into a case with infinite loop. But with the change of implementation with refactored structure it could look much safer and more straightforward.

modeling/local-slice-algorithm.go
ckyiricky: yeah, that makes sense. I added the tie-breaker as we discussed.

ckyiricky:
[Update]
Refactor to make balancing more straightforward
Add tie-breaker for unbalanced deviations

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,171 @@
...
+// the deviation from the expected number of endpoints in a perfectly balanced
+// system.
+// 3. EndpointSlices in zones with less endpoints than expected will receive
+// endpoints from zones that have more endpoints than expected.
+type LocalSliceAlgorithm struct{}
+
+type endpointDeviation struct {
+	name      string
+	deviation int
+}
```
robscott: Nit: this could use a comment explaining what deviation means in this case

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,171 @@
...
+
+// CreateSliceGroups creates sliceGroups with 'one local EndpointSliceGroup per
+// zone' policy
+func (alg LocalSliceAlgorithm) CreateSliceGroups(region regionInfo) (map[string]EndpointSliceGroup, error) {
+	if region.zoneDetails == nil {
+		return nil, fmt.Errorf("zoneDetail should not be nil")
+	}
+	sliceGroups := map[string]EndpointSliceGroup{}
+	// endpointsAvailable stores 'endpointDeviation' -- zones with 'extra'
+	// endpoints
```
robscott: Nit: Not sure you need to mention `endpointDeviation` here - I think the comment here could mirror the pattern you used below for `endpointsNeeded`.

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,171 @@
...
+		// this local sliceGroup should only receive traffic from current zone,
+		// this map tracks weights of traffic from different zones to this
+		// sliceGroup
+		localGroup.Weights = map[string]float64{zoneName: 1.0}
+		// this map stores composition of this sliceGroup, zoneName - number of
+		// endpoints from zoneName
+		localGroup.Composition = map[string]int{}
+
+		// the expected number of endpoints is with the same ratio as the number
+		// of nodes in this zone
```
robscott: Maybe `Calculate the expected number endpoints based on the proportion of nodes in this zone.`

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,171 @@
...
+		expectedEndpoints := int(zone.nodesRatio * float64(region.totalEndpoints))
+		// deviation: - means need more endpoints from other sliceGroups, +
+		// means need give out endpoints to other sliceGroups
```
robscott: The first time I read this it wasn't clear to me what `-` and `+` symbolized here, maybe you could replace those with "a positive value means..." and "a negative value means..."

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,171 @@
...
+
+	err := balanceSliceGroups(&endpointsAvailable, &endpointsNeeded, sliceGroups)
+	if err != nil {
+		return nil, err
+	}
+	return sliceGroups, nil
+}
+
+// balanceSliceGroups helps to 'rent' endpoints from zones have extra endpoints
+// to EndpointSliceGroups that have insufficient ones
```
robscott: Maybe `balanceSliceGroups distributes endpoints from zones with extra endpoints to EndpointSliceGroups for zones with insufficient endpoints`.

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,171 @@
...
+		// (calculated above) ends up in asymmetric sums of endpoints
+		// available and endpoints in need (in need > available)
+		if len(endpointsAvailable.byZone) == 0 {
+			makeupForApproximation(endpointsNeeded, sliceGroups, false)
+			return nil
+		}
+		// This is ugly, but the index below is a copy, after pop no way to
+		// udpate the index to its new position
+		indexDelta := 0
+		for index, sendZone := range endpointsAvailable.byZone {
```
robscott: In this case it might be simpler to use a traditional for loop where you manually control/increment i instead of having to maintain this separate value.

modeling/local-slice-algorithm.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// this helper function helps to deal with situation where the endpoints in need
+// != endpoints available due to the deviation caused by approximation while
+// calculating the expected endpoints of every zone.
+// 1. endpoints needed > endpoints available -> availableOverflow == false,
+// endpoints = list of needed endpoints
+// 2. endpoints available > endpoints needed -> availableOverflow == true,
+// endpoints = list of endpoints available
+func makeupForApproximation(endpoints *endpointsList, sliceGroups map[string]EndpointSliceGroup, availableOverflow bool) {
```
robscott: If I'm understanding this function correctly, I think we have some room to simplify it. As far as I can tell, it's called in 2 places right now:
1. Endpoints are needed but none are available - noop
2. Endpoints are available but none are needed - simple loop

I think we can get rid of 1 entirely which leaves us with 2. Given that 2 is only ~4 lines of code, I'd be tempted to just embed that into the function above around line 145.

robscott (commented):
I like where this is headed! Although we don't need full test coverage yet, it would likely be good to introduce some basic test coverage for these algorithms just to be sure we're getting the expected EndpointSliceGroups with a few different sets of input. That can come in a follow up PR, but that's likely the only way we'll know for sure that this is doing what we want it to.

modeling/local-slice-algorithm.go
ckyiricky: done

modeling/local-slice-algorithm.go
ckyiricky: makes sense, change to traditional loop

modeling/local-slice-algorithm.go
ckyiricky: done

modeling/local-slice-algorithm.go
ckyiricky: fixed

modeling/local-slice-algorithm.go
ckyiricky: fixed

modeling/local-slice-algorithm.go
ckyiricky: changed

modeling/local-slice-algorithm.go
ckyiricky: added

ckyiricky:
[Update]
Refine some comments/codes to be clearer
Embedded unnecessary helper functions

robscott (approved):
Thanks for the updates, looks good to me. Mind creating a follow up issue to add tests for this algorithm?

PR #9, opened 2020-08-03
Score mechanism for algorithms along with input/output with files

ckyiricky:
[New]
Add Input-Simulation-Output process with 3 goroutings
Add score mechanism for algorithms
Add algorithm factory
[Update]
Refactored data structures

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,33 +16,269 @@ limitations under the License.
 
 package main
 
-import "github.com/googleinterns/k8s-topology-simulator/modeling"
+import (
+	"encoding/csv"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/googleinterns/k8s-topology-simulator/modeling"
+)
 
 func main() {
-	zoneA := modeling.Zone{Nodes: 30, Endpoints: 60, Name: "a"}
-	zoneB := modeling.Zone{Nodes: 35, Endpoints: 70, Name: "b"}
-	zoneC := modeling.Zone{Nodes: 50, Endpoints: 80, Name: "c"}
-	zones := []modeling.Zone{zoneA, zoneB, zoneC}
+	// algorithm name, default shared global
+	algPtr := flag.String("alg", "SharedGlobalAlgorithm", "routing algorithm")
+	// input file
+	inputPtr := flag.String("input", "example/input.csv", "inputs to use for this algorithm")
+	// output file, default alg_result.csv
+	outputPtr := flag.String("output", "example/output.csv", "output of this algorithm")
+	flag.Parse()
 
-	alg, err := modeling.CreateAlg(0.4, 100)
+	var errList []<-chan error
+
+	zoneNames, inputQueue, errC, err := parseInput(*inputPtr)
+	errorHandler(err)
+	errList = append(errList, errC)
+
+	outputQueue, errC, err := startSimulation(*algPtr, inputQueue)
+	errorHandler(err)
+	errList = append(errList, errC)
+
+	errC, err = parseResult(*outputPtr, zoneNames, outputQueue)
+	errorHandler(err)
+	errList = append(errList, errC)
+
+	// wait for goroutines to return
+	err = waitForSimulation(errList...)
+	errorHandler(err)
+}
+
+func errorHandler(err error) {
 	if err != nil {
-		panic(err)
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+}
+
+// every row of the input file will be parsed to one instance of inputData
+type inputData struct {
+	// input id of the row
+	name string
+	// parse zone info of the input file into zone data structure
+	zones []modeling.Zone
+}
+
+// every instance of inputData will be mapped to one instance of outputData
+type outputData struct {
+	// same id as input id
+	name string
+	// simulation result of that piece of input data
+	result modeling.SimulationResult
+}
+
+// paserInput parses input csv file to instances of inputData and puts them into
+// a queue
+func parseInput(file string) ([]string, <-chan inputData, <-chan error, error) {
+	inputFile, err := os.Open(filepath.Join("", filepath.Clean(file)))
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	fmt.Printf("Reading data from %v\n", file)
+	reader := csv.NewReader(inputFile)
+	reader.TrimLeadingSpace = true
+	line, err := reader.Read()
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	var zoneNames []string
+	for _, name := range line[1:] {
+		name = strings.TrimSpace(name)
+		zoneNames = append(zoneNames, name)
+	}
+	dataC := make(chan inputData)
+	errC := make(chan error, 1)
+
+	go func() {
+		defer close(dataC)
+		defer close(errC)
+		for {
+			// read one row
+			rowCells, err := reader.Read()
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				errC <- err
+				return
+			}
+			if len(rowCells) != len(zoneNames)+1 {
+				fmt.Printf("[WARNING] unmatched number of fields, skip data %v\n", rowCells)
+				continue
+			}
+			var rowData inputData
+			rowData.name = rowCells[0]
+			for index, data := range rowCells[1:] {
+				nodestr := strings.Fields(data)
+				// convert string to int. number of nodes in a zone
+				numNode, err := strconv.Atoi(nodestr[0])
+				if err != nil {
+					errC <- err
+					return
+				}
+				// convert string to int. number of endpoints in a zone
+				numEndpoints, err := strconv.Atoi(nodestr[1])
+				if err != nil {
+					errC <- err
+					return
+				}
+				rowData.zones = append(rowData.zones, modeling.Zone{
+					Nodes:     numNode,
+					Endpoints: numEndpoints,
+					Name:      zoneNames[index],
+				})
+			}
+			dataC <- rowData
+		}
+	}()
+
+	return zoneNames, dataC, errC, nil
+}
+
+// startSimulation processes simulation on input data, produces instances of
+// outputData structure and puts them in a queue
+func startSimulation(algName string, inputQueue <-chan inputData) (<-chan outputData, <-chan error, error) {
+	// create algrithm based on the algorithm name delivered by the flag
+	alg := modeling.NewAlgorithm(algName)
+	// create simulation model, currently do calculation based on probability
+	// rather than real simulation.
+	model, err := modeling.NewModel(alg, modeling.TheoreticalSimulator{})
+	if err != nil {
+		return nil, nil, err
 	}
 
-	var sim modeling.TheoreticalSimulator
+	resultC := make(chan outputData)
+	errC := make(chan error, 1)
 
-	// TODO: operations on metrics rather than printing only
-	model, err := modeling.NewModel(zones, alg, sim)
+	go func() {
+		defer close(resultC)
+		defer close(errC)
+		for {
+			rowData, more := <-inputQueue
+			if !more {
+				break
+			}
+			err := model.UpdateRegion(rowData.zones)
+			if err != nil {
+				errC <- fmt.Errorf("input id : %s, %v", rowData.name, err)
+				return
+			}
+			simRes, err := model.StartSimulation()
+			if err != nil {
+				errC <- fmt.Errorf("input id : %s, %v", rowData.name, err)
+				return
+			}
+			var result outputData
+			result.name = rowData.name
+			result.result = simRes
+			resultC <- result
+		}
+	}()
+
+	return resultC, errC, nil
+}
+
+// parseResult parses outputData to evaluation metrics and writes back to a
+// result file
+func parseResult(file string, zoneNames []string, outputQueue <-chan outputData) (<-chan error, error) {
+	outputFile, err := os.Create(file)
 	if err != nil {
-		panic(err)
+		return nil, err
+	}
+	defer func() {
+		cerr := outputFile.Close()
+		if err == nil {
+			err = cerr
+		}
+	}()
+	fmt.Printf("Creating output to file %v\n", file)
+	writer := csv.NewWriter(outputFile)
+
+	title := []string{"input id", "score", "in-zone-traffic score", "deviation score", "max deviation", "mean deviation", "SD of deviation"}
+	err = writer.Write(title)
+	if err != nil {
+		return nil, err
+	}
+	errC := make(chan error, 1)
+	defer close(errC)
+	for {
+		rowData, more := <-outputQueue
+		if !more {
+			break
+		}
+
+		// use in zone traffic percentage to be in zone traffic score
+		inZoneTrafficScore := rowData.result.InZoneTraffic * 100
+		// use mean deviation to calcualte deviation score
+		deviationScore := 100.0 - rowData.result.MeanDeviation*100
+		// calculate total score based on two scores above
+		totalScore := 0.6*inZoneTrafficScore + 0.4*deviationScore
+
+		data := []string{rowData.name}
+		data = append(data, strconv.FormatFloat(totalScore, 'f', 4, 64))
+		data = append(data, strconv.FormatFloat(inZoneTrafficScore, 'f', 4, 64))
+		data = append(data, strconv.FormatFloat(deviationScore, 'f', 4, 64))
+		data = append(data, strconv.FormatFloat(rowData.result.MaxDeviation*100, 'f', 4, 64)+"%")
+		data = append(data, strconv.FormatFloat(rowData.result.MeanDeviation*100, 'f', 4, 64)+"%")
+		data = append(data, strconv.FormatFloat(rowData.result.DeviationSD, 'f', 4, 64))
+
+		err = writer.Write(data)
+		if err != nil {
+			return errC, err
+		}
+	}
+	writer.Flush()
+	err = writer.Error()
+	if err != nil {
+		return errC, err
+	}
+	return errC, nil
+}
+
+// waitForSimulation waits for above routings return and handle their errors
+func waitForSimulation(errs ...<-chan error) error {
+	errC := mergeErrors(errs...)
+	for err := range errC {
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// mergeErrors handles errors in the goroutines
+func mergeErrors(errs ...<-chan error) <-chan error {
+	var wg sync.WaitGroup
+	outC := make(chan error, len(errs))
+	output := func(c <-chan error) {
+		for err := range c {
+			outC <- err
+		}
+		wg.Done()
+	}
+	wg.Add(len(errs))
+
+	for _, errC := range errs {
+		go output(errC)
 	}
 
-	err = model.StartSimulation()
-	if err != nil {
-		panic(err)
-	}
-	err = model.PrintLastResult()
-	if err != nil {
-		panic(err)
-	}
+	go func() {
+		wg.Wait()
+		close(outC)
+	}()
+	return outC
 }
--- a/main.go
+++ b/main.go
@@ -25,9 +25,9 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync"
 
 	"github.com/googleinterns/k8s-topology-simulator/modeling"
+	"k8s.io/klog/v2"
 )
 
 func main() {
@@ -38,29 +38,21 @@ func main() {
 	// output file, default alg_result.csv
 	outputPtr := flag.String("output", "example/output.csv", "output of this algorithm")
 	flag.Parse()
+	klog.InitFlags(nil)
 
-	var errList []<-chan error
-
-	zoneNames, inputQueue, errC, err := parseInput(*inputPtr)
+	inputArray, err := parseInput(*inputPtr)
 	errorHandler(err)
-	errList = append(errList, errC)
 
-	outputQueue, errC, err := startSimulation(*algPtr, inputQueue)
+	outputArray, err := startSimulation(*algPtr, inputArray)
 	errorHandler(err)
-	errList = append(errList, errC)
 
-	errC, err = parseResult(*outputPtr, zoneNames, outputQueue)
-	errorHandler(err)
-	errList = append(errList, errC)
-
-	// wait for goroutines to return
-	err = waitForSimulation(errList...)
+	err = parseResult(*outputPtr, outputArray)
 	errorHandler(err)
 }
 
 func errorHandler(err error) {
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "%v\n", err)
+		klog.Errorf("%v\n", err)
 		os.Exit(1)
 	}
 }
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
 
 func main() {
-	zoneA := modeling.Zone{Nodes: 30, Endpoints: 60, Name: "a"}
-	zoneB := modeling.Zone{Nodes: 35, Endpoints: 70, Name: "b"}
-	zoneC := modeling.Zone{Nodes: 50, Endpoints: 80, Name: "c"}
-	zones := []modeling.Zone{zoneA, zoneB, zoneC}
+	// algorithm name, default shared global
+	algPtr := flag.String("alg", "SharedGlobalAlgorithm", "routing algorithm")
+	// input file
+	inputPtr := flag.String("input", "test.csv", "info of each zone")
```
robscott: Maybe create an `examples` folder that includes sample input and output?
```suggestion
	inputPtr := flag.String("input", "examples/input.csv", "Inputs to use for this algorithm")
```

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	"github.com/googleinterns/k8s-topology-simulator/modeling"
+)
 
 func main() {
-	zoneA := modeling.Zone{Nodes: 30, Endpoints: 60, Name: "a"}
-	zoneB := modeling.Zone{Nodes: 35, Endpoints: 70, Name: "b"}
-	zoneC := modeling.Zone{Nodes: 50, Endpoints: 80, Name: "c"}
-	zones := []modeling.Zone{zoneA, zoneB, zoneC}
+	// algorithm name, default shared global
+	algPtr := flag.String("alg", "SharedGlobalAlgorithm", "routing algorithm")
```
robscott, line 35: You might want to look into using [Cobra](https://github.com/spf13/cobra) here for some more advanced functionality. That could include more detail for each flag and an automatic `--help` output.

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	// input file
+	inputPtr := flag.String("input", "test.csv", "info of each zone")
+	// output file, default alg_result.csv
+	outputPtr := flag.String("output", "", "simulation results")
```
robscott: I think it might be better to default to a more generic name here.
```suggestion
	outputPtr := flag.String("output", "output.csv", "simulation results")
```

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	flag.Parse()
+	if *outputPtr == "" {
+		*outputPtr = *algPtr + "_result.csv"
+	}
```
robscott: Lots of potential for nil pointer exceptions. I think it would be simpler just to have a consistent output.csv default.

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
 
-	alg, err := modeling.CreateAlg(0.4, 100)
+	var errList []<-chan error
```
robscott: This feels strange - why do we need an error list here?

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+
+	zoneNames, inputQueue, errC, err := parseInput(*inputPtr)
 	if err != nil {
 		panic(err)
```
robscott: I know this isn't new, but I'd recommend avoiding the use of `panic()`. This describes a more common pattern I've seen used: https://stackoverflow.com/questions/18963984/exit-with-error-code-in-go.

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
 	}
-
-	var sim modeling.TheoreticalSimulator
-
-	// TODO: operations on metrics rather than printing only
-	model, err := modeling.NewModel(zones, alg, sim)
+	errList = append(errList, errC)
+	outputQueue, errC, err := startSimulation(*algPtr, inputQueue)
```
robscott: I think the extra error channel returned by this function is not necessary here.

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
 	if err != nil {
 		panic(err)
 	}
-
-	err = model.StartSimulation()
+	errList = append(errList, errC)
+	errC, err = parseResult(*outputPtr, zoneNames, outputQueue)
```
robscott: similarly not sure we need to make use of an error channel here.

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
 	if err != nil {
 		panic(err)
 	}
-	err = model.PrintLastResult()
+	errList = append(errList, errC)
+	// wait for goroutings to return
```
robscott: Nit: goroutines

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	err = waitForSimulation(errList...)
 	if err != nil {
 		panic(err)
 	}
 }
+
+// every row of the input file will be parsed to one instance of inputData
+type inputData struct {
+	// input id of the row
+	id string
```
robscott: Maybe `name` would be better than `id` now that this field's a string?

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	// parse zone info of the input file into zone data structure
+	zones []modeling.Zone
+}
+
+// every piece of input data will be processed to one instance of outputData
```
robscott: Nit: maybe a couple tiny changes:
```suggestion
// every instance of inputData will be mapped to one instance of outputData
```

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+type outputData struct {
+	// same id as input id
+	id string
+	// simulation result of that piece of input data
+	result modeling.SimulationResult
+}
+
+// paserInput parses input csv file to instances of inputData and puts them into
+// a queue
+func parseInput(file string) ([]string, <-chan inputData, <-chan error, error) {
```
robscott: There's a lot going on in `main.go` now. It would likely make sense to split some of this out into separate files.

robscott: The signature of this function seems more complicated than it ideally would be. Can we just make this return `[]string, error`?

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	inputFile, err := os.Open(filepath.Join("", filepath.Clean(file)))
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	fmt.Printf("Reading data from %v\n", file)
```
robscott: Not necessary in this PR, but it might be good to use a logging library here like [glog](https://github.com/golang/glog) or [logrus](https://github.com/sirupsen/logrus).

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	}
+	dataC := make(chan inputData)
+	errC := make(chan error, 1)
+
+	go func() {
+		defer close(dataC)
+		defer close(errC)
+		for {
+			// read one row
+			datas, err := reader.Read()
```
robscott: Maybe `rowCells` instead of `datas`

robscott: Everything below this could probably be split out into a function that parses a row and returns `inputData, err`. Then these can be added to a channel after the function returns.

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	var zoneNames []string
+	for _, name := range line[1:] {
+		name = strings.TrimSpace(name)
+		zoneNames = append(zoneNames, name)
+	}
+	dataC := make(chan inputData)
+	errC := make(chan error, 1)
```
robscott: You may want to make use of wait groups here such as https://golangcode.com/errors-in-waitgroups/.

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+			}
+			if err != nil {
+				errC <- err
+				return
+			}
+			if len(datas) != len(zoneNames)+1 {
+				fmt.Printf("[WARNING] unmatched number of fields, skip data %v\n", datas)
+				continue
+			}
+			var oneData inputData
```
robscott: Maybe `rowData`?

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+					errC <- err
+					return
+				}
+				// convert string to int. number of endpoints in a zone
+				numEndpoints, err := strconv.Atoi(nodestr[1])
+				if err != nil {
+					errC <- err
+					return
+				}
+				oneData.zones = append(oneData.zones, modeling.Zone{Nodes: numNode, Endpoints: numEndpoints, Name: zoneNames[index]})
```
robscott: This line feels a bit long, maybe this would be better?
```suggestion
				oneData.zones = append(oneData.zones, modeling.Zone{
					Nodes: numNode,
					Endpoints: numEndpoints,
					Name: zoneNames[index],
				})
```

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+
+	go func() {
+		defer close(resultC)
+		defer close(errC)
+		for {
+			oneData, more := <-inputQueue
+			if !more {
+				break
+			}
+			err := model.UpdateRegion(oneData.zones)
```
robscott: Is this threadsafe?

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	}
+
+	resultC := make(chan outputData)
+	errC := make(chan error, 1)
+
+	go func() {
+		defer close(resultC)
+		defer close(errC)
+		for {
+			oneData, more := <-inputQueue
```
robscott: I think everything in this loop could likely be extracted to a separate function. This would also be easier to unit test.

main.go
```diff
@@ -16,33 +16,269 @@ limitations under the License.
...
+	errC := mergeErrors(errs...)
+	for err := range errC {
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// mergeErrors handles errors in the goroutings
```
robscott: Nit: goroutines

modeling/algorithm-factory.go
```diff
@@ -0,0 +1,35 @@
...
+*/
+
+package modeling
+
+import (
+	"fmt"
+)
+
+// MakeAlgorithm serves as an algorithm constructor based on the algroithm name
+func MakeAlgorithm(name string) (RoutingAlgorithm, error) {
```
robscott: Even though factory is a common term, I haven't actually seen anything commonly called that in Go (referring to filename here). I'm more used to a function like this being called `NewAlgorithm`. I'm also not used to seeing functions like this return an error. Maybe you could ditch the `CreatedSharedGlobalAlg` func and just add that logic here? No need to validate input if it's hardcoded.

modeling/data.go
```diff
@@ -51,42 +42,96 @@ type EndpointSliceGroup struct {
...
+	ZoneTrafficWeights map[string]float64
 }
 
-// Traffic represents the detailed traffic infomation of a zone
-type Traffic struct {
+// SimulationResult is to collect metrics of a simulation result
+type SimulationResult struct {
+	// InZoneTraffic is the total ratio of traffic that stays in the same zone
+	InZoneTraffic float64
+	// TrafficDistribution groups Traffic by zone name
```
robscott: Nit: `zoneTraffic`

modeling/data.go
```diff
@@ -51,42 +42,96 @@ type EndpointSliceGroup struct {
...
+	// Total number of endpoints of all zones
+	totalEndpoints int
+	// Detailed information of each zone, zone name - exact zone
+	zoneDetails map[string]Zone
+}
+
+// endpoints with weights are used to do routing inside an EndpointSliceGroup
+type weightedEndpoints struct {
+	// Number of endpoints
+	Number int
```
robscott: Does this field need to be exported? Seems unlikely given that the enclosing type is not exported.

modeling/data.go
```diff
@@ -51,42 +42,96 @@ type EndpointSliceGroup struct {
...
+	// Weights of these endpoints when routing in a slice
+	Weight float64
```
robscott: Same question about exporting

modeling/data.go
```diff
@@ -51,42 +42,96 @@ type EndpointSliceGroup struct {
...
+}
+
+// zoneTraffic records the detailed traffic infomation of a zone
+type zoneTraffic struct {
```
robscott, line 83: Same questions about exporting for all the below fields

modeling/local-slice-algorithm.go
```diff
@@ -146,7 +146,8 @@ func balanceSliceGroups(endpointsAvailable *endpointsList, endpointsNeeded *endp
 		// in this case, we assign those extra endpoints to its local
 		// endpointSliceGroups
 		for _, extraEndpoints := range endpointsAvailable.byZone {
-			sliceGroups[extraEndpoints.name].Composition[extraEndpoints.name] += extraEndpoints.deviation
+			originalEndpoints := sliceGroups[extraEndpoints.name].Composition[extraEndpoints.name]
+			sliceGroups[extraEndpoints.name].Composition[extraEndpoints.name] = weightedEndpoints{Number: originalEndpoints.Number + extraEndpoints.deviation, Weight: originalEndpoints.Weight}
```
robscott: This is a long line, maybe try splitting it up a bit:
```suggestion
			sliceGroups[extraEndpoints.name].Composition[extraEndpoints.name] = weightedEndpoints{
				Number: originalEndpoints.Number + extraEndpoints.deviation,
				Weight: originalEndpoints.Weight,
			}
```

modeling/model.go
```diff
@@ -18,98 +18,73 @@ package modeling
...
 			totalSlices++
 		}
 	}
 	return totalSlices
 }
+
+// GetLastResult gets the last result of the simulation
+func (m *Model) GetLastResult() SimulationResult {
+	return m.results[len(m.results)-1]
+}
```
robscott: Is this still being used? I thought we were aggregating results + printing out each individual result?

modeling/theoretical-simulator.go
```diff
@@ -27,48 +30,142 @@ func (sim TheoreticalSimulator) Simulate(region regionInfo, endpointSlices map[s
...
-				kubeProxy[name][compZone] += float64(pods) * slice.Weights[name]
+	// weighted number of endpoints in each sliceGroup can be reached by every
+	// zone
+	// first key: zone name
+	// second key: sliceGroup label
+	zoneReachableEndpointsInSG := make(map[string]map[string]float64)
+	for zone := range region.zoneDetails {
+		zoneReachableEndpointsInSG[zone] = map[string]float64{}
+		for sliceLabel, slice := range endpointSlices {
+			zoneReachableEndpointsInSG[zone][sliceLabel] = float64(slice.numberOfEndpoints()) * slice.ZoneTrafficWeights[zone]
```
robscott: What if `slice.ZoneTrafficWeights[zone]` is not set?

modeling/theoretical-simulator.go
```diff
@@ -27,48 +30,142 @@ func (sim TheoreticalSimulator) Simulate(region regionInfo, endpointSlices map[s
...
+	// total ratio of traffic received by each EndpointSliceGroup
+	sgTrafficRatio := map[string]float64{}
+	for label := range endpointSlices {
+		for zone := range region.zoneDetails {
+			sgTrafficRatio[label] += region.zoneDetails[zone].nodesRatio * zoneTrafficToSG[zone][label]
+		}
+	}
+	// endpoints traffic load
+	// first key: zone the endpoints belong to
+	// second key: sliceGroup the endpoints assigned to
```
robscott: I think this pattern is used 3x in this function - maybe we could use a struct that contained all this information for a sliceGroup instead of 3 separate maps?

modeling/theoretical-simulator.go
```diff
@@ -27,48 +30,142 @@ func (sim TheoreticalSimulator) Simulate(region regionInfo, endpointSlices map[s
...
+	endpointsTrafficLoad := map[string]map[string]float64{}
+	for zone := range region.zoneDetails {
```
robscott: I think there are also 3 loops through `region.zoneDetails` in this function, is there a way we could merge some of these?

modeling/theoretical-simulator.go
```diff
@@ -27,48 +30,142 @@ func (sim TheoreticalSimulator) Simulate(region regionInfo, endpointSlices map[s
...
 		}
 	}
 
 	// calculate result of one simulation
 	var simResult SimulationResult
-	simResult.TrafficDetail = map[string]Traffic{}
-	simResult.Workload = map[string]float64{}
-	for zoneName, zone := range region.zoneDetails {
+	// traffic distribution details by zone
+	simResult.TrafficDistribution = map[string]zoneTraffic{}
```
robscott, line 151: This has turned into a really long function, one that would be rather difficult to effectively unit test. I think there are lots of opportunities to split this up, but this is one of the big ones - calculating the results of the simulation.

modeling/data.go
robscott: Related: The `modeling` package is getting quite large. It may make sense to split out some parts like simulation into different packages. That doesn't necessarily have to happen yet, but if you were already about to, you could export any types that might be needed by other packages.

robscott (commented):
Lots of great progress here, thanks! This is a huge PR and I'm sure I've missed things but here's a first round of review.

main.go
ckyiricky: fixed

main.go
ckyiricky: fixed

main.go
ckyiricky: fixed

main.go
ckyiricky: Got it, this looks much better!

main.go
ckyiricky: fixed

main.go
ckyiricky: changed

main.go
ckyiricky: changed

main.go
ckyiricky: todo in future update

main.go
ckyiricky: changed

main.go
ckyiricky: changed

main.go
ckyiricky: fixed

modeling/algorithm-factory.go
ckyiricky: changed and renamed the file to algorithms.go

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed. That makes sense, could do this after this PR, it should involve some work to split the package reasonably

modeling/data.go
ckyiricky: fixed

modeling/data.go
ckyiricky: fixed

modeling/local-slice-algorithm.go
ckyiricky: changed

modeling/theoretical-simulator.go
ckyiricky: if not set, deem as 0. if its a mistake, this should not be the responsibility for this function to take. The creator should take care of the weights value and set it correctly.

modeling/theoretical-simulator.go
ckyiricky: Done

modeling/theoretical-simulator.go
ckyiricky: Done

modeling/theoretical-simulator.go
ckyiricky: Done

ckyiricky:
[TODO]
Going to change the goroutines in main.go as discussed in the next update.

[Update]
Code cleaning for theoretical-simulator.go
Typo fixed
Minor changed

modeling/model.go
ckyiricky: changed the func signature that currently the StartSimulation will return the result. no need to call this separately.

ckyiricky:
[Update]
Split inputParse-simulation-outputParse into serial processes
Add klog

main.go
```diff
@@ -16,33 +16,206 @@ limitations under the License.
...
+	errorHandler(err)
+
+	outputArray, err := startSimulation(*algPtr, inputArray)
+	errorHandler(err)
+
+	err = parseResult(*outputPtr, outputArray)
+	errorHandler(err)
+}
+
+func errorHandler(err error) {
```
robscott: Maybe rename this to `exitWithError()`

main.go
```diff
@@ -16,33 +16,206 @@ limitations under the License.
...
+	model, err := modeling.NewModel(alg, modeling.TheoreticalSimulator{})
 	if err != nil {
-		panic(err)
+		return nil, err
+	}
+	var outputArray []outputData
+	for _, rowData := range inputArray {
+		err := model.UpdateRegion(rowData.zones)
+		if err != nil {
+			return outputArray, fmt.Errorf("input name : %s, %v", rowData.name, err)
```
robscott:
```suggestion
			return outputArray, fmt.Errorf("error updating region for input %s, %v", rowData.name, err)
```

main.go
```diff
@@ -16,33 +16,206 @@ limitations under the License.
...
+		}
+		simRes, err := model.StartSimulation()
+		if err != nil {
+			return outputArray, fmt.Errorf("input name : %s, %v", rowData.name, err)
```
robscott:
```suggestion
			return outputArray, fmt.Errorf("error starting simulation for input %s, %v", rowData.name, err)
```

main.go
```diff
@@ -16,33 +16,206 @@ limitations under the License.
...
+		}
+		var result outputData
+		result.name = rowData.name
+		result.result = simRes
+		outputArray = append(outputArray, result)
```
robscott: Would this work?
```suggestion
		outputArray = append(outputArray, outputData{name: rowData.name, result: simRes})
```

main.go
```diff
@@ -16,33 +16,206 @@ limitations under the License.
...
 	}
-	err = model.PrintLastResult()
+	defer func() {
+		cerr := outputFile.Close()
+		if err == nil {
+			err = cerr
+		}
+	}()
+
+	klog.Infof("Creating output to file %v\n", file)
```
robscott:
```suggestion
	klog.Infof("Writing output to file %v\n", file)
```

main.go
```diff
@@ -16,33 +16,206 @@ limitations under the License.
...
+		outputArray = append(outputArray, result)
 	}
 
-	err = model.StartSimulation()
+	return outputArray, nil
+}
+
+// parseResult parses outputData to evaluation metrics and writes back to a
+// result file
+func parseResult(file string, outputArray []outputData) (err error) {
```
robscott, line 173: There's a lot in `main.go`, ideally this file should be pretty minimal and call out to other packages. Not needed in this PR, but as we start to think about splitting this into multiple packages, I think we'll also want to move a lot of the content here out into a new package.

modeling/data.go
```diff
@@ -51,42 +42,96 @@ type EndpointSliceGroup struct {
...
-	// of capacity
-	Workload map[string]float64
+	// TrafficDistribution groups zoneTraffic by zone name
+	TrafficDistribution map[string]zoneTraffic
+	// MaxDeviation of traffic load of all endpoints
+	MaxDeviation float64
+	// MeanDeviation of traffic load of all endpoints
+	MeanDeviation float64
+	// DeviationSD abbr of standard deviation of the daviation of traffic load
+	// of all endpoints
```
robscott:
```suggestion
	// DeviationSD represents the standard deviation of the deviation of traffic load
	// across all endpoints
```

modeling/shared-global-algorithm.go
```diff
@@ -60,42 +60,34 @@ func (alg SharedGlobalAlgorithm) CreateSliceGroups(region regionInfo) (map[strin
...
 		// If total pods <= threshold, all pods go to global slice
 		if region.totalEndpoints <= alg.globalThreshold {
-			globalEndpoints = zone.Endpoints
+			globalEndpoints.number = zone.Endpoints
+			globalEndpoints.weight = 1
 		} else {
 			// Otherwise calculate the global contribution of current zone based
 			// on the global weight and the deviation of this zone
-			globalEndpoints = int(math.Min(math.Max(0.0, deviation[name])/alg.globalWeight, float64(zone.Endpoints)))
+			globalEndpoints.number = int(math.Min(math.Max(0.0, deviation[name])/alg.globalWeight, float64(zone.Endpoints)))
```
robscott, line 77: There's a lot of math going on here. Maybe add a comment to explain it and/or split it out into a couple lines.

modeling/theoretical-simulator.go
```diff
@@ -27,48 +30,174 @@ func (sim TheoreticalSimulator) Simulate(region regionInfo, endpointSlices map[s
...
+func (zd zoneSGDetails) getEndpointsTrafficLoadDetails(region regionInfo, endpointSlices map[string]EndpointSliceGroup) {
+	// total ratio of traffic received by each EndpointSliceGroup
+	sgTrafficRatio := map[string]float64{}
+	for zone, sgDetails := range zd {
+		for label, trafficRatio := range sgDetails.zoneTrafficRatio {
+			sgTrafficRatio[label] += region.zoneDetails[zone].nodesRatio * trafficRatio
+		}
+	}
+
+	// theoratially, traffic should be distributed equally among all the
```
robscott:
```suggestion
	// theoretically, traffic should be distributed equally among all the
```

modeling/theoretical-simulator.go
```diff
@@ -27,48 +30,174 @@ func (sim TheoreticalSimulator) Simulate(region regionInfo, endpointSlices map[s
...
+	// endpoints
+	theoretialTrafficLoad := 1.0 / float64(region.totalEndpoints)
```
robscott:
```suggestion
	theoreticalTrafficLoad := 1.0 / float64(region.totalEndpoints)
```

robscott (commented):
Thanks for the work on this! Just a few minor nits.

main.go
ckyiricky: Agreed, I was about to split these functions to other packages, but I think we might do the package split with the `modeling` as well. So I kept the main.go this way, going to create an issue for this TODO.

main.go
ckyiricky: Done

main.go
ckyiricky: Done

main.go
ckyiricky: Done

main.go
ckyiricky: Looks great. Changed

main.go
ckyiricky: Done

modeling/data.go
ckyiricky: Done

modeling/shared-global-algorithm.go
ckyiricky: Added comments to explain

modeling/theoretical-simulator.go
ckyiricky: fixed

modeling/theoretical-simulator.go
ckyiricky: fixed

ckyiricky:
[Update]
Bug fix : allow zones to have 0 endpoints/nodes
Typo fix

robscott (approved):
LGTM, thanks!

PR #12, opened 2020-08-10
[Refactor] Split package modeling into smaller packages

ckyiricky:
Split package modeling into smaller packages
    Move data to package data
    Move algorithms to package algorithm
    Move simulator to package simulator
Split package main into smaller packages
    Move input/output parse to package process
Update Readme to match new packages

[New] Add original algorithm
Add original algorithm to be benchmark


process/input-parser.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if err != nil {
+		return inputData{}, true, err
+	}
+	var rowData inputData
+	rowData.name = rowCells[0]
+	for index, data := range rowCells[1:] {
+		nodestr := strings.Fields(data)
+		// convert string to int. number of nodes in a zone
+		numNode, err := strconv.Atoi(nodestr[0])
```
robscott: Nit: numNodes

modeling/data/data.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package modeling
+package data
```
robscott: Although I'd usually recommend keeping the types in the same package as they're primarily used, the ones in this package mostly seem to be used by multiple packages. Assuming there's no good way to get away from a types specific package, maybe this should be renamed `types` instead of `data`?

modeling/data/data.go
```diff
@@ -62,104 +62,105 @@ type SimulationResult struct {
...
-		total += float64(endpoints.number) * endpoints.weight
+		total += float64(endpoints.Number) * endpoints.Weight
 	}
 	return total
 }
 
-// Helper function to create regionInfo with zone infos
-func createRegionInfo(zones []Zone) (regionInfo, error) {
+// CreateRegionInfo creates regionInfo with zone infos
+func CreateRegionInfo(zones []Zone) (RegionInfo, error) {
```
robscott: I know this is not actually new here, so this doesn't have to be fixed here, but I think it might make more sense to change this to `func NewRegionInfo(zones []Zone) RegionInfo {`. Instead of returning an error, you could just return empty `RegionInfo` and leave it up to the caller to determine if that's valid or not.

robscott (commented):
Thanks for the work on this!

process/input-parser.go
```diff
@@ -0,0 +1,97 @@
...
+	if err == io.EOF {
+		return inputData{}, true, nil
+	}
+	if err != nil {
+		return inputData{}, true, err
+	}
+	var rowData inputData
+	rowData.name = rowCells[0]
+	for index, data := range rowCells[1:] {
+		nodestr := strings.Fields(data)
```
robscott: Nit: nodeStr

process/input-parser.go
ckyiricky: Fixed

modeling/data/data.go
ckyiricky: Renamed to types.

process/input-parser.go
ckyiricky: Fixed

robscott (approved):
LGTM, thanks!

PR #15, opened 2020-08-14
[Add] Variations of local and shared algorithms

ckyiricky:
[Add] Add two variations of local-slice-algorithm
    1. Handle corner cases with a shared slice.
    2. Use weights to distribute 'float' endpoints
[Update] Score weights change
    Change weights for different metrics
    Add mean deviation and max deviation both into consideration


modeling/algorithm/algorithms.go
```diff
@@ -27,6 +27,12 @@ func NewAlgorithm(name string) RoutingAlgorithm {
 	case "Local", "LocalAlgorithm", "LocalSliceAlgorithm":
 		klog.Info("LocalSliceAlgorithm created")
 		return LocalSliceAlgorithm{}
+	case "LocalWeighted":
+		klog.Info("LocalWeightedSliceAlgorithm created")
+		return LocalWeightedSliceAlgorithm{}
+	case "LocalOpt", "LocalAlgorithmOpt", "LocalSliceAlgorithmOpt":
```
robscott: Maybe `LocalShared` instead to match the type.

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,254 @@
...
+	"math"
+
+	"github.com/googleinterns/k8s-topology-simulator/modeling/types"
+	"k8s.io/klog/v2"
+)
+
+// LocalSharedSliceAlgorithm is one variation of LocalSliceAlgorithm which
+// 'borrows' and 'rents' endpoints from other zones to make the local
+// EndpointSlice balanced with the incoming traffic (number of nodes
+// distribution). This variation deals with failed corner cases while having a
```
robscott: How does it deal with failed corner cases?

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,254 @@
...
+// slightly different distribution policy when approximating float to int
+type LocalSharedSliceAlgorithm struct{}
+
+// CreateSliceGroups creates sliceGroups with 'one local EndpointSliceGroup per
+// zone' policy, zones with 0.xx expected endpoints while have no endpoints in
+// zone will be treated with a higher priority.
```
robscott: Maybe something like this?
```suggestion
// zone' policy, zones with no endpoints allocated will be treated with a higher priority.
```

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,254 @@
...
+func (alg LocalSharedSliceAlgorithm) CreateSliceGroups(region types.RegionInfo) (map[string]types.EndpointSliceGroup, error) {
+	if region.ZoneDetails == nil {
+		return nil, fmt.Errorf("zoneDetail should not be nil")
+	}
+	// if number of total endpoints < number of zones, use original algorithm
+	// instead. This algorithm itself can handle some of these special corner
+	// cases but performs poorly in most of the time, so using the original
```
robscott:
```suggestion
	// cases but performs poorly at larger scale, so using the original
```

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,254 @@
...
+		return OriginalAlgorithm{}.CreateSliceGroups(region)
+	}
+	sliceGroups := map[string]types.EndpointSliceGroup{}
+	// endpointsAvailable stores zones with number of endpoints available
+	endpointsAvailable := endpointsList{}
+	// endpointsNeeded stores zones with number of endpoints needed
+	endpointsNeeded := endpointsList{}
+	// endpointsNeeded stores high priority zones with number of endpoints
+	// needed
+	endpointsNeededUrgent := endpointsList{}
```
robscott:
```suggestion
	// endpointsNeededUrgent stores high priority zones with number of endpoints
	// needed
	endpointsNeededUrgent := endpointsList{}
```

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,254 @@
...
+			localGroup.Composition[zoneName] = types.WeightedEndpoints{Number: zone.Endpoints, Weight: 1}
+		}
+		sliceGroups[zoneName] = localGroup
+	}
+	urgentPool.SliceGroups = sliceGroups
+	receiverPool.SliceGroups = sliceGroups
+	heap.Init(&urgentPool)
+	heap.Init(&receiverPool)
+
+	succ, err := alg.balanceSliceGroups(&endpointsAvailable, &endpointsNeeded, &endpointsNeededUrgent, sliceGroups, &urgentPool, &receiverPool)
```
robscott: Since we know that we don't have to deal with scenarios where the number of endpoints is less than the number of zones, maybe we can simplify this a bit. This is likely an oversimplification, but it seems like we want to treat the case of zones with no local endpoints as a special one. In that case, we would simply have a single slice group that contained `max(sum(expectedEndpoints), 1)` endpoints and was distributed to all zones that didn't have local endpoints. The rest would then be a relatively simple assignment problem. Admittedly I'm likely missing some of the edge cases here.

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return sliceGroups, nil
+}
+
+// balanceSliceGroups distributes endpoints from zones with extra endpoints to
+// EndpointSliceGroups for zones with insufficient endpoints. It serves
+// high-priority zones first.
+func (alg LocalSharedSliceAlgorithm) balanceSliceGroups(endpointsAvailable *endpointsList, endpointsNeeded *endpointsList, endpointsNeededUrgent *endpointsList, sliceGroups map[string]types.EndpointSliceGroup, urgentPool *ZonePriorityQueue, receiverPool *ZonePriorityQueue) (bool, error) {
```
robscott: I think we may be able to simplify this to avoid the need for urgent special cases.

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,202 @@
...
+		// this map stores composition of this sliceGroup, zoneName - number of
+		// endpoints from zoneName
+		localGroup.Composition = map[string]types.WeightedEndpoints{}
+
+		// calculate expected number of endpoints based on the proportion of
+		// nodes in this zone
+		expectedEndpoints := zone.NodesRatio * float64(region.TotalEndpoints)
+		// deviation: a negative value means need more endpoints from other
+		// sliceGroups, a positive value means need give out endpoints to other
+		// sliceGroups
```
robscott:
```suggestion
		// deviation: a negative value means this zone needs more endpoints from other
		// zones, a positive value means this zone needs to give out endpoints to other
		// zones
```

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,202 @@
...
+			localGroup.Composition[zoneName] = types.WeightedEndpoints{Number: zone.Endpoints, Weight: 1}
+		}
+		sliceGroups[zoneName] = localGroup
+
+		// push 0.xx deviation into corresponding lists
+		floatDeviation := deviation - float64(intDeviation)
+		if floatDeviation > 0 {
+			// Intentionally set deviation to default 0, this means the zone
+			// itself needs float part from this endpoint otherwise means the
+			// endpoints will be shared by other zones only
```
robscott: This comment is a bit confusing to me

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,202 @@
...
+			weightedEndpointsAvailable.push(endpointDeviation{name: zoneName, weight: floatDeviation})
+		} else if floatDeviation < 0 {
+			weightedEndpointsNeeded.push(endpointDeviation{name: zoneName, deviation: 1, weight: -floatDeviation})
```
robscott: Why is deviation 1 here? Why is weight -floatDeviation?

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// balanceSliceGroups distributes endpoints from zones with extra endpoints to
+// EndpointSliceGroups for zones with insufficient endpoints.
+func (alg LocalWeightedSliceAlgorithm) balanceSliceGroups(endpointsAvailable *endpointsList, endpointsNeeded *endpointsList, weightedEndpointsAvailable *endpointsList, weightedEndpointsNeeded *endpointsList, sliceGroups map[string]types.EndpointSliceGroup) error {
+
+	for _, receiveZone := range endpointsNeeded.byZone {
+		// No more 'int' endpoints available but are still 'int' endpoints in
+		// need, we push the needed entpoints into weighted list and deal them
+		// as float endpoints
```
robscott: Not quite sure I'm understanding this correctly, but maybe something like:
```
There are no more full endpoints available, but this zone still needs endpoints. Push the needed endpoints into a weighted list and deal with them as partial endpoints.
```

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,202 @@
...
+			}
+			if sendZone.deviation < receiveZone.deviation {
+				sliceGroups[receiveZone.name].Composition[sendZone.name] = types.WeightedEndpoints{Number: sendZone.deviation, Weight: 1}
+				receiveZone.deviation -= sendZone.deviation
+				endpointsAvailable.pop()
+				continue
+			}
+		}
+		// if needed.deviation != 0 means more 'int' needed than available, push
+		// to weighted list and deal as float
```
robscott: Isn't this similar to the comment on 103?

robscott (commented):
Thanks for the work on this! Didn't quite make it through everything for this first round, but had some higher level questions around structure.

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: yes, but this is an edge case that the '103' can't deal with. Simply its one deviation has remaining number needed but no more available endpoints to serve it, we have to push that deviation into float list and treat as float endpoints.

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: done

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: Done

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: changed

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: Done

modeling/algorithm/local-shared-algorithm.go
ckyiricky: done

modeling/algorithm/local-shared-algorithm.go
ckyiricky: changed

modeling/algorithm/local-shared-algorithm.go
ckyiricky: Added more comments about it

modeling/algorithm/algorithms.go
ckyiricky: Fixed, sorry for the silly mismatch here.

modeling/algorithm/local-shared-algorithm.go
ckyiricky: change to more specific cases

ckyiricky:
[Add/Update] Add/Updated three local variations
    [Add]
    Add one local variation that assigns the extra endpoints available at
    the end to a global sliceGroup which has evenly distributed routing
    weights to all zones
    [Update]
    Update local-weighted algorithms with clearer code and comments
    Update local-shared with simpler implementation
        Treat urgent zones as a whole
        Use priority queue to assign endpoints to zones need endpoints
    [Add/Update]
        Update input
        Update alg outputs and add new output for new algs

ckyiricky:
Add a shared global variation which excludes contributors from the global slices

example/topology-sim-input.csv
```diff
@@ -5,6 +5,7 @@ name,zone1,zone2,zone3
 balanced minimal endpoints,30 1,30 1,30 1
 unbalanced minimal endpoints,30 2,30 1,30 0
 only 1 endpoint,30 1,30 0,30 0
+only 2 endpoint,30 2,30 0,30 0
```
robscott, line 8: I think it would be good to also add one or more of the edge cases we'd discussed earlier. Can't quite remember exactly what they were, but I know there was one something like this:

```
a b c
2 3, 0 1, 0 1 
```

modeling/algorithm/algorithms.go
```diff
@@ -23,10 +23,22 @@ func NewAlgorithm(name string) RoutingAlgorithm {
 	switch name {
 	case "SharedGlobal", "SharedGlobalAlgorithm":
 		klog.Info("SharedGlobalAlgorithm created")
-		return SharedGlobalAlgorithm{globalWeight: 0.4, globalThreshold: 100}
+		return SharedGlobalAlgorithm{sharedCoreAlgorithm: SharedGlobalAlgorithmCore{globalWeight: 0.4, globalThreshold: 100}}
+	case "SharedGlobalExclude", "SharedGlobalAlgorithmExclude":
+		klog.Info("SharedGlobalAlgorithmExclude created")
```
robscott: I think this should by `SharedMultiZone`

modeling/algorithm/algorithms.go
```diff
@@ -23,10 +23,22 @@ func NewAlgorithm(name string) RoutingAlgorithm {
...
 	case "Local", "LocalAlgorithm", "LocalSliceAlgorithm":
 		klog.Info("LocalSliceAlgorithm created")
 		return LocalSliceAlgorithm{}
+	case "LocalWeighted":
+		klog.Info("LocalWeightedSliceAlgorithm created")
+		return LocalWeightedSliceAlgorithm{}
+	case "LocalOpt", "LocalAlgorithmOpt", "LocalSliceAlgorithmOpt":
+		klog.Info("LocalSliceAlgorithmOpt created")
+		return LocalSliceAlgorithmOpt{}
+	case "LocalShared":
```
robscott: Nit: For some algorithms you have multiple acceptable name and for others you don't. I think for consistency it would be best to support a single name per algorithm.

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+	"fmt"
+	"math"
+
+	"github.com/googleinterns/k8s-topology-simulator/modeling/types"
+	"k8s.io/klog/v2"
+)
+
+// LocalSharedSliceAlgorithm is one variation of LocalSliceAlgorithm which
+// 'borrows' and 'rents' endpoints from other zones to make the local
+// EndpointSlice balanced with the incoming traffic (number of nodes
```
robscott: Nit: EndpointSliceGroup, EndpointSlices or similar to indicate it isn't just a single EndpointSlice.

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+// distribution). This variation deals with failed corner cases by sharing
+// endpoints to zones has no endpoints.
```
robscott: Nit: that have no endpoints

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+	if region.ZoneDetails == nil {
+		return nil, fmt.Errorf("zoneDetail should not be nil")
+	}
+	// if number of total endpoints < number of zones, use original algorithm
+	// instead. This algorithm itself can handle some of these special corner
+	// cases but performs poorly at small scale corner cases, so using the
+	// original algorithm seems a better solution in terms of performance and
+	// simplicity.
+	if region.TotalEndpoints < len(region.ZoneDetails) {
+		return OriginalAlgorithm{}.CreateSliceGroups(region)
```
robscott, line 48: I forget, is the OrignalAlgorithm just the global/shared one? If so, maybe we should change the name here to make it more clear what it does.

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	sliceGroups := map[string]types.EndpointSliceGroup{}
+	// endpointsNeeded stores zones with number of endpoints needed
+	endpointsNeeded := endpointsList{}
+	// endpointsNeededUrgent stores high priority zones with number of endpoints
```
robscott, line 53: Maybe explain what high priority zones are here

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+		// sliceGroups
+		deviation := float64(zone.Endpoints) - expectedEndpoints
+		// if deviation > 0, this zone is a qualified candidate for
+		// availablePool that contributes endpoints to other zones
+		validCandidate := !math.Signbit(deviation)
+		if validCandidate {
+			availablePool.ZoneNames = append(availablePool.ZoneNames, zoneName)
+		}
+		receiverPool.ZoneNames = append(receiverPool.ZoneNames, zoneName)
+		// merge all the zones have no endpoints into a whole
```
robscott: Maybe something like this?
```suggestion
		// merge all the zones with no endpoints into a shared slice group
```

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+		if zone.Endpoints == 0 {
+			// this is a form used to accuratly represent the deviation in
```
robscott:
```suggestion
			// this is a form used to accurately represent the deviation in
```

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+			// float, floatDeviation = 1 * -deviation
```
robscott: This comment is confusing, did you mean to remove it?

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+		if deviation <= -1 {
+			endpointsNeeded.push(endpointDeviation{name: zoneName, deviation: int(-deviation)})
+		}
+	}
+	availablePool.SliceGroups = sliceGroups
+	receiverPool.SliceGroups = sliceGroups
+	heap.Init(&availablePool)
+	heap.Init(&receiverPool)
+
+	succ, err := alg.balanceSliceGroups(&endpointsNeeded, &endpointsNeededUrgent, sliceGroups, &availablePool, &receiverPool)
```
robscott: Not quite sure I understand this, why do availablePool and receiverPool get equivalent slice groups assigned to them? Can any of this logic be moved into balanceSliceGroups?

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+	if !succ {
+		klog.Infoln("fail to do local algorithm, switch to shared global algorithm")
```
robscott:
```suggestion
		klog.Infoln("failed to use local algorithm, switching to shared global algorithm")
```

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+	return sliceGroups, nil
+}
+
+// balanceSliceGroups distributes endpoints from zones with extra endpoints to
+// EndpointSliceGroups for zones with insufficient endpoints.
+func (alg LocalSharedSliceAlgorithm) balanceSliceGroups(endpointsNeeded *endpointsList, endpointsNeededUrgent *endpointsList, sliceGroups map[string]types.EndpointSliceGroup, availablePool *ZonePriorityQueue, receiverPool *ZonePriorityQueue) (bool, error) {
+	// merge one sharedSG that zones in the urgent list will consume
+	mergedSG := types.EndpointSliceGroup{Composition: map[string]types.WeightedEndpoints{}, ZoneTrafficWeights: map[string]float64{}}
+	mergedED := endpointDeviation{name: "merged"}
+	mergedDeviation := 0.0
```
robscott: I know it's hard to name similar things here, but maybe a comment or something would help explain what the difference between these 2 deviations is.

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+	for _, urgentZone := range endpointsNeededUrgent.byZone {
+		mergedED.name += "-" + urgentZone.name
+		mergedDeviation += (float64(urgentZone.deviation) * urgentZone.weight)
+		mergedSG.ZoneTrafficWeights[urgentZone.name] = 1
+		endpointsNeededUrgent.pop()
+	}
+	// take the ceil, in case the deviation = 0.xx, we have to make sure there
```
robscott: Would this also work?
```suggestion
	// take the ceil, if the deviation > 0, we have to make sure there
```

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+	// the second and third zone will not require endpoints based on floor
+	// approximation (2.8 -> 2). But the 1st zone has too many endpoints, it
+	// should give one out.
+	for {
+		if availablePool.Len() == 0 {
+			break
+		}
+		candidate := heap.Pop(availablePool).(string)
+		expectedEndpoints := float64(availablePool.Region.TotalEndpoints) * availablePool.Region.ZoneDetails[candidate].NodesRatio
+		if float64(sliceGroups[candidate].Composition[candidate].Number)-expectedEndpoints >= 1 {
```
robscott: There's a lot going on this if statement - could use some nil checks. Maybe this could be a helper func.

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+	// i.e. (nodes, endpoints), (1 3, 2 2, 2 2)
+	// the second and third zone will not require endpoints based on floor
+	// approximation (2.8 -> 2). But the 1st zone has too many endpoints, it
+	// should give one out.
+	for {
+		if availablePool.Len() == 0 {
+			break
+		}
+		candidate := heap.Pop(availablePool).(string)
+		expectedEndpoints := float64(availablePool.Region.TotalEndpoints) * availablePool.Region.ZoneDetails[candidate].NodesRatio
```
robscott: This could use a nil check, maybe a helper func would be good here

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+		if float64(sliceGroups[candidate].Composition[candidate].Number)-expectedEndpoints >= 1 {
+			receiveZone := heap.Pop(receiverPool).(string)
+			updateSGComposition(sliceGroups[receiveZone], candidate, 1, 1)
+			heap.Push(receiverPool, receiveZone)
+
+			updateSGComposition(sliceGroups[candidate], candidate, -1, 1)
+			heap.Push(availablePool, candidate)
```
robscott: There's a lot of key logic here, some comments to explain what is happening in each step would be helpful

modeling/algorithm/local-shared-algorithm.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+		break
```
robscott: This continue and break combination along with an otherwise infinite loop feels a bit strange to me, is there another way we could structure this loop?

modeling/algorithm/local-slice-algorithm-opt.go
```diff
@@ -0,0 +1,136 @@
...
+		// sliceGroups
+		deviation := float64(zone.Endpoints) - expectedEndpoints
+		if deviation > 0 {
+			endpointsAvailable.push(endpointDeviation{name: zoneName, deviation: int(math.Ceil(deviation))})
+			localGroup.Composition[zoneName] = types.WeightedEndpoints{Number: int(expectedEndpoints), Weight: 1}
+		} else if deviation < 0 {
+			endpointsNeeded.push(endpointDeviation{name: zoneName, deviation: int(-deviation)})
+			localGroup.Composition[zoneName] = types.WeightedEndpoints{Number: int(zone.Endpoints), Weight: 1}
+		} else {
+			localGroup.Composition[zoneName] = types.WeightedEndpoints{Number: int(zone.Endpoints), Weight: 1}
```
robscott: This seems to be almost completely repeated in each of the three clauses above, maybe initialize the weightedEndpoints above and only change the number in the one case it's necessary.

modeling/algorithm/local-slice-algorithm-opt.go
```diff
@@ -0,0 +1,136 @@
...
+			if sendZone.deviation > receiveZone.deviation {
+				sliceGroups[receiveZone.name].Composition[sendZone.name] = types.WeightedEndpoints{Number: receiveZone.deviation, Weight: 1}
+				endpointsAvailable.byZone[index].deviation -= receiveZone.deviation
+				break
+			}
+			if sendZone.deviation < receiveZone.deviation {
+				sliceGroups[receiveZone.name].Composition[sendZone.name] = types.WeightedEndpoints{Number: sendZone.deviation, Weight: 1}
+				receiveZone.deviation -= sendZone.deviation
+				endpointsAvailable.pop()
+				continue
```
robscott: Do we need continue here?

modeling/algorithm/local-slice-algorithm-opt.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+		}
+		endpointsNeeded.pop()
+	}
+	// all endpoints should have been distributed properly. This happens when
+	// the sum of approximated available endpoints > sum of approximated
+	// endpoints in need
+	if len(endpointsAvailable.byZone) != 0 {
+		// in this case, we assign those extra endpoints to a global
+		// endpointSliceGroups
```
robscott:
```suggestion
		// endpointSliceGroup
```

modeling/algorithm/local-slice-algorithm-opt.go
```diff
@@ -0,0 +1,136 @@
...
+				receiveZone.deviation -= sendZone.deviation
+				endpointsAvailable.pop()
+				continue
+			}
+		}
+		endpointsNeeded.pop()
+	}
+	// all endpoints should have been distributed properly. This happens when
+	// the sum of approximated available endpoints > sum of approximated
+	// endpoints in need
```
robscott: These two sentences seem to contradict each other. If all endpoints had been distributed at this point it seems like we wouldn't need this section.

modeling/algorithm/local-slice-types.go
```diff
@@ -0,0 +1,118 @@
+/*
```
robscott: This seems to include more than just local slice types, maybe a more generic utils.go name would be a better fit here.

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+		localGroup.Composition = map[string]types.WeightedEndpoints{}
+
+		// calculate expected number of endpoints based on the proportion of
+		// nodes in this zone
+		expectedEndpoints := zone.NodesRatio * float64(region.TotalEndpoints)
+		// deviation: a negative value means this zone needs more endpoints from
+		// other zones, a positive value means this zone needs to give out
+		// endpoints to other zones
+		deviation := float64(zone.Endpoints) - expectedEndpoints
+		// intDeviation is dealt same way as original local-slice algorithm that
```
robscott:
```suggestion
		// intDeviation is dealt with the same way as original local-slice algorithm that
```

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+		// directly 'borrows' or 'sends' endpoints among zones
+		intDeviation := int(deviation)
+		if intDeviation == 0 {
+			localGroup.Composition[zoneName] = types.WeightedEndpoints{Number: int(expectedEndpoints), Weight: 1}
+		} else if intDeviation > 0 {
+			endpointsAvailable.push(endpointDeviation{name: zoneName, deviation: intDeviation})
+			localGroup.Composition[zoneName] = types.WeightedEndpoints{Number: int(expectedEndpoints), Weight: 1}
+		} else {
+			endpointsNeeded.push(endpointDeviation{name: zoneName, deviation: -intDeviation})
+			localGroup.Composition[zoneName] = types.WeightedEndpoints{Number: zone.Endpoints, Weight: 1}
```
robscott: I had a similar comment above that these WeightedEndpoints could likely be declared above the if statements to reduce the duplication.

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+		sliceGroups[zoneName] = localGroup
+
+		// push 0.xx deviation into corresponding lists
+		// 2.3 expected endpoints, 3 ownd endpoints, 0.7 decimal deviation, 0.3
+		// endpoints and 2 endpoints needed
+		// 2.4 expected endpoints, 1 owned endpoint, -0.4 decimal deviation, 0.4
+		// endpoints and 2 endpoints needed
+		decimalDeviation := deviation - float64(intDeviation)
+		if decimalDeviation > 0 {
+			// One endpoint from this zone will contribute 1-decimalDeviation to
```
robscott, line 92: What's the logic behind this equation, maybe add a quick explanation of why this equation is used here as a comment.

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+			// the local zone, and contribute remaining to other zones. This
+			// will be implemented through routing weights of EndpointSliceGroup
+			weightedEndpointsAvailable.push(endpointDeviation{name: zoneName, deviation: 1, weight: 1 - decimalDeviation, consumeByLocal: true})
+		} else if decimalDeviation < 0 {
+			weightedEndpointsNeeded.push(endpointDeviation{name: zoneName, deviation: 1, weight: -decimalDeviation})
```
robscott, line 99: Maybe add a comment here as well to explain why weight is being set to -decimalDeviation.

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+	}
+
+	err := alg.balanceSliceGroups(&endpointsAvailable, &endpointsNeeded, &weightedEndpointsAvailable, &weightedEndpointsNeeded, sliceGroups)
+	return sliceGroups, err
+}
+
+// balanceSliceGroups distributes endpoints from zones with extra endpoints to
+// EndpointSliceGroups for zones with insufficient endpoints.
+func (alg LocalWeightedSliceAlgorithm) balanceSliceGroups(endpointsAvailable *endpointsList, endpointsNeeded *endpointsList, weightedEndpointsAvailable *endpointsList, weightedEndpointsNeeded *endpointsList, sliceGroups map[string]types.EndpointSliceGroup) error {
```
robscott, line 109: How similar is this to the other balanceSliceGroups functions? Is there a way we could share some code here between algorithms?

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+		// There are no more full endpoints available, but this zone still needs
+		// endpoints. Push the needed endpoints into a weighted list and deal
+		// with them as partial endpoints.
+		if len(endpointsAvailable.byZone) == 0 {
+			receiveZone.weight = 1
+			weightedEndpointsNeeded.push(receiveZone)
+			endpointsNeeded.pop()
+			continue
+		}
+		// same logic as original local-slice algorithm
```
robscott: If this is shared logic maybe we can extract it to a helper func?

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+			if sendZone.deviation < receiveZone.deviation {
+				sliceGroups[receiveZone.name].Composition[sendZone.name] = types.WeightedEndpoints{Number: sendZone.deviation, Weight: 1}
+				receiveZone.deviation -= sendZone.deviation
+				endpointsAvailable.pop()
+				continue
+			}
+		}
+		// if needed.deviation != 0 means more full endpoints needed than
+		// available, push to weighted list and deal them as partial endpoints
```
robscott:
```suggestion
		// available, push to weighted list and deal with them as partial endpoints
```

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+				break
+			}
+			if sendZone.deviation < receiveZone.deviation {
+				sliceGroups[receiveZone.name].Composition[sendZone.name] = types.WeightedEndpoints{Number: sendZone.deviation, Weight: 1}
+				receiveZone.deviation -= sendZone.deviation
+				endpointsAvailable.pop()
+				continue
+			}
+		}
+		// if needed.deviation != 0 means more full endpoints needed than
```
robscott: When would this happen? It seems strange to equate != 0 meaning more needed than available, wouldn't a negative value have a different meaning than a positive value here?

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	// distribute 'float' endpoints with weights. create shared slices among
+	// zones that ends up with sum(weights for each zone in the SG) = 1
+	// use weights to implement float endpoints for a zone, i.e. 0.4 endpoints
+	// for zoneA will be implemented with a SG having 1 endpoint but 0.4 routing
+	// weight to zoneA
+	for _, extraEndpoints := range weightedEndpointsAvailable.byZone {
+		sharedSlice := types.EndpointSliceGroup{Label: "shared", Composition: map[string]types.WeightedEndpoints{}, ZoneTrafficWeights: map[string]float64{}}
+		// If this endpoint will be consumbed by its local zone, contribue to
```
robscott:
```suggestion
		// If this endpoint will be consumed by its local zone, contribute to
```

modeling/algorithm/local-weighted-slice-algorithm.go
```diff
@@ -0,0 +1,211 @@
...
+				break
+			}
+			// needed < available, update the available value and serve the next
+			// needed one
+			if deviation < 0 {
+				sharedSlice.ZoneTrafficWeights[receiveZone.name] += float64(receiveZone.deviation) * receiveZone.weight / float64(extraEndpoints.deviation)
+				sharedSlice.Label += "-" + receiveZone.name
+				extraEndpoints.weight -= sharedSlice.ZoneTrafficWeights[receiveZone.name]
+				weightedEndpointsNeeded.pop()
+				continue
```
robscott: Wouldn't this happen automatically?

modeling/algorithm/shared-global-algorithm-core.go
```diff
@@ -0,0 +1,101 @@
...
+	// Threshold of global EndpointSliceGroup that if the total number of endpoints
+	// <= threshold, all endpoints go to global EndpointSliceGroup
+	globalThreshold int
+}
+
+// CreateSliceGroups takes a region of zones as input and output
+// EndpointSliceGroups
+func (alg SharedGlobalAlgorithmCore) CreateSliceGroups(region types.RegionInfo, excludeContributor bool) (map[string]types.EndpointSliceGroup, error) {
+	if region.ZoneDetails == nil {
+		return nil, errors.New("can't create EndpointSlices with 0 number of zone")
```
robscott:
```suggestion
		return nil, errors.New("can't create EndpointSlices without zones specified")
```

modeling/algorithm/shared-global-algorithm-core.go
```diff
@@ -0,0 +1,101 @@
...
+
+	for _, zone := range region.ZoneDetails {
+		// Calculate the deviation based on the capacity(endpoints) and
+		// traffic(nodes) ratio
+		deviation[zone.Name] = float64(zone.Endpoints) - float64(region.TotalEndpoints)*zone.NodesRatio
+	}
+
+	// Output EndpointSlices
+	sliceGroups := make(map[string]types.EndpointSliceGroup)
+	// The global sliceGroup -- might be split into many small global slices
```
robscott: Had a similar question above, what does `--` mean here?

modeling/algorithm/shared-global-algorithm-core.go
```diff
@@ -0,0 +1,101 @@
...
+	var globalSliceGroup types.EndpointSliceGroup
+	globalSliceGroup.Label = "global"
+	globalSliceGroup.Composition = make(map[string]types.WeightedEndpoints)
+	globalSliceGroup.ZoneTrafficWeights = make(map[string]float64)
+	for name, zone := range region.ZoneDetails {
+		var globalEndpoints types.WeightedEndpoints
+		// If total pods <= threshold, all pods go to global slice
+		if region.TotalEndpoints <= alg.globalThreshold {
+			globalEndpoints.Number = zone.Endpoints
+			globalEndpoints.Weight = 1
```
robscott: Elsewhere you're just using the original algorithm in similar cases, would that make sense here as well?

robscott (commented):
Thanks for all your work on this, lots of good stuff in here! Really appreciate all the comments throughout to help explain the logic. One overarching comment is that it seems like there's a lot of duplicate code between these algorithms. I know relying on shared functions could overcomplicate changes to a specific algorithm in the future, but maybe there's some room to share some code that's less likely to change.

modeling/algorithm/algorithms.go
ckyiricky: changed

modeling/algorithm/algorithms.go
ckyiricky: That makes sense, I changed to two names for every algorithms : xxx and xxxAlgorithm

modeling/algorithm/local-shared-algorithm.go
ckyiricky: fixed

modeling/algorithm/local-shared-algorithm.go
ckyiricky: fixed

modeling/algorithm/local-shared-algorithm.go
ckyiricky: original algorithms if the  'evenly distributed' policy.

modeling/algorithm/local-shared-algorithm.go
ckyiricky: added

modeling/algorithm/local-shared-algorithm.go
ckyiricky: changed

modeling/algorithm/local-shared-algorithm.go
ckyiricky: fixed

modeling/algorithm/local-shared-algorithm.go
ckyiricky: add more comments to explain the move here

modeling/algorithm/local-shared-algorithm.go
ckyiricky: move into balanceSliceGroups

modeling/algorithm/local-shared-algorithm.go
ckyiricky: fixed

modeling/algorithm/local-shared-algorithm.go
ckyiricky: added

modeling/algorithm/local-shared-algorithm.go
ckyiricky: changed

modeling/algorithm/local-shared-algorithm.go
ckyiricky: added a helper function

modeling/algorithm/local-shared-algorithm.go
ckyiricky: added a nil detection inside the helper function. In this case, it should be guaranteed to be non-nil.

modeling/algorithm/local-shared-algorithm.go
ckyiricky: added

modeling/algorithm/local-shared-algorithm.go
ckyiricky: restructured

modeling/algorithm/local-slice-algorithm-opt.go
ckyiricky: changed

modeling/algorithm/local-slice-algorithm-opt.go
ckyiricky: deleted

modeling/algorithm/local-slice-algorithm-opt.go
ckyiricky: fixed

modeling/algorithm/local-slice-algorithm-opt.go
ckyiricky: Deleted the first sentence

modeling/algorithm/local-slice-types.go
ckyiricky: renamed

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: fixed

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: changed

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: added

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: added

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: Created a helper function to share some common logic

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: wrapped the logic to a helper function, same as the previous comment

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: fixed

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: changed to > 0 and added more comments to explain

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: fixed

modeling/algorithm/local-weighted-slice-algorithm.go
ckyiricky: deleted

modeling/algorithm/shared-global-algorithm-core.go
ckyiricky: changed

modeling/algorithm/shared-global-algorithm-core.go
ckyiricky: sorry for the legacy comment here, updated to a more reasonable one

modeling/algorithm/shared-global-algorithm-core.go
ckyiricky: sorry about the legacy issue here, updated to using original algorithm

ckyiricky:
[Update] variation update
    Update local* algorithms with fewer duplicated codes
    Rename shared-global-algorithm-exclude to shared-multizone-algorithm
    Fix typos

example/topology-sim-input.csv
ckyiricky: Created an issue for the corner cases update

robscott (approved):
Thanks for all your work on this!

